PR #308, opened 2020-12-18
Process link regex matches through JS for validation

timszot:
@marcaaron (PB) and @alex-mechler  will you please review this?

Updated our `ExpensiMark` to capture URL links more reliably.

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/147205

# Tests
The `ExpensiMark` tests have been updated to test various URL linking.

# QA
NA


lib/ExpensiMark.js
```diff
@@ -137,4 +149,46 @@ export default class ExpensiMark {
...
+
+            if (abort) {
+                replacedText = replacedText.concat(textToCheck.substr(match.index, (match[0].length)));
+            } else {
+                replacedText = replacedText.concat(replacement(match[0], match[1], match[2], match[3]));
+            }
+            startIndex = match.index + (match[0].length);
+            match = regex.exec(textToCheck);
+        }
+        if (startIndex < textToCheck.length) replacedText = replacedText.concat(textToCheck.substr(startIndex));
```
alex-mechler, line 190: I _think_ we try to avoid 1 line ifs like this

lib/ExpensiMark.js
```diff
@@ -137,4 +149,46 @@ export default class ExpensiMark {
...
+    /**
+     * Checks matched URLs for validity and replace valid links with html elements
+     *
+     * @param {RegExp} regex
+     * @param {String} textToCheck
+     * @param {Function} replacement
+     *
+     * @returns {String}
+     */
+    modifyTextForUrlLinks(regex, textToCheck, replacement) {
```
alex-mechler, line 162: This function is specifically written for our URL regex, so IMO we should not be passing it here.

lib/ExpensiMark.js
```diff
@@ -137,4 +149,46 @@ export default class ExpensiMark {
...
+
+            replacedText = replacedText.concat(textToCheck.substr(startIndex, (match.index - startIndex)));
+
+            if (abort) {
+                replacedText = replacedText.concat(textToCheck.substr(match.index, (match[0].length)));
+            } else {
+                replacedText = replacedText.concat(replacement(match[0], match[1], match[2], match[3]));
+            }
+            startIndex = match.index + (match[0].length);
+            match = regex.exec(textToCheck);
```
alex-mechler, line 188: NAB: a small comment about how this moves to the next match would be good. Its not obvious if you don't understand how JS's regex engine runs.

lib/ExpensiMark.js
```diff
@@ -54,6 +58,8 @@ export default class ExpensiMark {
                 name: 'link',
                 regex: new RegExp(`\\[([\\w\\s\\d!?&#;]+)\\]\\(${URL_REGEX}\\)(?![^<]*(<\\/pre>|<\\/code>))`, 'gi'),
```
alex-mechler, line 59: It feels weird to have both "regex" and "process" defined here, yet are only calling the process. I would recommend either moving the regex definition into the function, or into the arrow function in "process".

alex-mechler (changes requested):
Can you also fix the warns so they dont appear at the bottom of other prs? (like combobox is doing in this one)

lib/ExpensiMark.js
timszot: The linter had me change it to this.

lib/ExpensiMark.js
timszot: The regex is not the same since this is used for both the `link` and `autolink` rules. That's why it's being passed to the function.

lib/ExpensiMark.js
alex-mechler: Ah, if the linter wants it, the linter wants it!

lib/ExpensiMark.js
alex-mechler: Ah yes, forgot those two were using differennt regexes. This is good then!

timszot:
Updated!

lib/ExpensiMark.js
```diff
@@ -1,7 +1,11 @@
 import Str from './str';
 import TLD_REGEX from './tlds';
 
-const URL_REGEX = `((https?:\\/\\/(?:www\\.)?)?([-a-z0-9]+?\\.)+(?:${TLD_REGEX})(?:(?:\\?|#|\\/)[-\\w()@:;%\\+.,~#&\\/={}\\[\\]?]*[-\\w()@:;%\\+~#?&\\/={}\\[\\]]|\\/|\\b|(?=_)))`;
+const URL_WEBSITE_REGEX = `(https?:\\/\\/)?((?:www\\.)?[-a-z0-9]+?\\.)+(?:${TLD_REGEX})(?:\\:\\d{2,4}|\\b|(?=_))`;
+const URL_PATH_REGEX = '(?:\\/[-\\w$@.&+!*"\'(),=%]*[-\\w~@:%)]|\\/)*';
+const URL_PARAM_REGEX = '\\?[-\\w$@.&+!*"\'(),=%{}:;\\[\\]]+';
+const URL_FRAGMENT_REGEX = '#[-\\w$@.&+!*"\'(),=%;\\/]*';
+const URL_REGEX = `(${URL_WEBSITE_REGEX}${URL_PATH_REGEX}(?:${URL_PARAM_REGEX}|${URL_FRAGMENT_REGEX})?)`;
```
alex-mechler, line 8: You can have a url with both a query param and a fragment, so I think this last bit should be 
`(?:${URL_PARAM_REGEX}?${URL_FRAGMENT_REGEX}?)

alex-mechler (changes requested):
Looks mostly good, found 1 failing case manually testing

timszot:
Updated again and added new URL to test suite.

alex-mechler (approved):
@marcaaron All you!

lib/ExpensiMark.js
marcaaron: Hmm what did it say exactly... maybe we can tell it what we want ?!

marcaaron (approved):
LGTM

PR #334, opened 2021-02-13
Rename .jsx files to .js and vice versa

nbhargava:

Optimistically, in making changes to the Expensify.cash desktop app, I want to avoid making webpack changes that require being able to parse JSX. `expensify-common` is not built or packed, so the imports that trace to it are made without any modifications. When the electron app tries to drive through the tree of pre-existing imports, it gets to a JSX file that causes problems.

Right now `CONST.jsx` is the main offender, but I figure it's worth switching some of the others. It is worth noting that there are still lint errors, but those seem to be independent of the changes I made. Mine just elucidate them.

### Fixed Issues
N/A

# Tests

Ran `npm lint` and saw no issues that were newly caused by renaming.

Ran `npm lint` and saw no issues that were newly caused by renaming.

# QA


nbhargava:
_**I have read the CLA Document and I hereby sign the CLA**_

roryabraham:
@nbhargava Is this PR still going to be necessary? I read [your comment here](https://github.com/Expensify/Expensify.cash/pull/1318#issuecomment-779335098) and it led me to think you might not need to do this anymore?

nbhargava:
Yeah probably not. Closing now.

PR #340, opened 2021-03-01
Add StagingDeployCash logic

AndrewGable:
@roryabraham  will you please review this?

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/155326

# Tests
1. Automated tests
2. If you want to see it not use mocks you can set `GITHUB_TOKEN` env variable and run this test (expect it to fail, but see real API results in failure)

```js
    test('Test finding an issue', () => {
        const octokit = new Octokit({
            auth: process.env.GITHUB_TOKEN
        });
        const github = new GithubUtils(octokit);
        return github.getStagingDeployCash().then(data => expect(data).toStrictEqual(expectedResponse));
    });
```

# QA
1. N/A

### DIFF
--- a/__tests__/GithubUtils-test.js
+++ b/__tests__/GithubUtils-test.js
@@ -29,15 +29,15 @@ describe('GithubUtils.getStagingDeployCash', () => {
         const octokit = new Octokit();
         const github = new GithubUtils(octokit);
         octokit.issues.listForRepo = jest.fn().mockResolvedValue({data: [{a: 1}, {b: 2}]});
-        return github.getStagingDeployCash().catch(e =>
-            expect(e).toEqual(new Error('Found more than one StagingDeployCash issue.')));
+        return github.getStagingDeployCash()
+            .catch(e => expect(e).toEqual(new Error('Found more than one StagingDeployCash issue.')));
     });
 
     test('Test finding no issues', () => {
         const octokit = new Octokit();
         const github = new GithubUtils(octokit);
         octokit.issues.listForRepo = jest.fn().mockResolvedValue({data: []});
-        return github.getStagingDeployCash().catch(e =>
-            expect(e).toEqual(new Error('Unable to find StagingDeployCash issue.')));
+        return github.getStagingDeployCash()
+            .catch(e => expect(e).toEqual(new Error('Unable to find StagingDeployCash issue.')));
     });
 });
--- a/__tests__/GithubUtils-test.js
+++ b/__tests__/GithubUtils-test.js
@@ -1,9 +1,9 @@
 import {Octokit} from '@octokit/rest';
 import GithubUtils from '../lib/GithubUtils';
 
-const expectedResponse = {
-    title: 'Andrew Test Issue',
+const issue = {
     url: 'https://api.github.com/repos/Andrew-Test-Org/Public-Test-Repo/issues/29',
+    title: 'Andrew Test Issue',
     labels: [
         {
             id: 2783847782,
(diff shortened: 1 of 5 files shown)


AndrewGable:
Draft until I get specifics on issue content: https://github.com/Expensify/Expensify/issues/155326#issuecomment-788337236

package.json
```diff
@@ -18,6 +18,7 @@
     "url": "git+ssh://git@github.com/Expensify/JS-Libs.git"
   },
   "dependencies": {
+    "@octokit/rest": "^18.3.1",
```
roryabraham, line 21: Curious if this needs to be a production dependency. Since the `octokit` client will be passed in from other places in code, and is just referenced here. I think the only place we're actually going to need to access the module itself is in tests, so maybe this could be a dev dependency.

AndrewGable: Good call! Let me try to make it a dev dependency.

lib/GithubUtils.js
```diff
@@ -0,0 +1,76 @@
+export const GithubOwner = 'Expensify';
```
roryabraham, line 1: NAB but because these are just constants they should probably be `UPPER_SNAKE_CASE`

lib/GithubUtils.js
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Takes in a GitHub issue object and returns the data we want
+     *
+     * @param {Object} issue
+     * @returns {Promise}
+     */
+    getStagingDeployCashData(issue) {
+        try {
+            const versionRegex = /`([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9]*))?`/g;
```
roryabraham, line 45:
```suggestion
            const versionRegex = /`([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9]+))?`/g;
```

lib/GithubUtils.js
```diff
@@ -0,0 +1,76 @@
...
+            const tag = issue.body.match(versionRegex)[0].replace(/`/g, '');
+
+            // eslint-disable-next-line max-len
+            const compareURLRegex = /https:\/\/github\.com\/Expensify\/Expensify.cash\/compare\/([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9]*))?...([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9]*))?/g;
```
roryabraham, line 49: I think if you declare `versionRegex` like `new RegExp('your_regex_here', 'g')`, then you can combine regexes. Here you would use it like this:

```javascript
const compareURLRegex = new RegExp(`https:\/\/github\.com\/Expensify\/Expensify\.cash\/compare\/${versionRegex.source}\.\.\.${versionRegex.source}`, 'g');
```

Also, you've got some unescaped `.` characters in here that I don't think you want (such as in `Expensify.cash` and for the `...`)

lib/GithubUtils.js
```diff
@@ -0,0 +1,76 @@
...
+            const comparisonURL = issue.body.match(compareURLRegex)[0];
+
+            const PRListRegex = /https:\/\/github\.com\/Expensify\/Expensify.cash\/pull\/([0-9]+)/g;
```
roryabraham, line 52:
```suggestion
            const PRListRegex = /https:\/\/github\.com\/Expensify\/Expensify\.cash\/pull\/([0-9]+)/g;
```

lib/GithubUtils.js
```diff
@@ -0,0 +1,76 @@
...
+            const PRList = issue.body.match(PRListRegex);
+
+            const data = {
+                title: issue.title, url: issue.url, labels: issue.labels, tag, comparisonURL, PRList
+            };
+
+            return this.octokit.git.getRef({
```
roryabraham, line 59: Hmmm I'm not sure this will work going forward to determine if the `StagingDeployCash` is deployed to production. But also, I'm not sure if we'll even need `isDeployedToProdction`, since in general any open `StagingDeployCash` _should_ be deployed to staging and any closed `StagingDeployCash` _should_ be deployed to production. The way to tell for sure would be to check if there is a release matching the tag you are looking for.

roryabraham (changes requested):

lib/GithubUtils.js
AndrewGable: Ok I am happy to either remove it or use `octokit.repos.getReleaseByTag`, just let me know which you'd prefer.

roryabraham: I looked through the design doc and it looks like we're not using `isDeployedToProduction` anywhere, so we should just remove it.

roryabraham (changes requested):
Overall LGTM, just need to escape `.` in your regexes.

__tests__/GithubUtils-test.js
```diff
@@ -0,0 +1,81 @@
...
+    title: 'Andrew Test Issue',
+    url: 'https://api.github.com/repos/Andrew-Test-Org/Public-Test-Repo/issues/29'
+};
+
+describe('GithubUtils.getStagingDeployCash', () => {
+    test('Test finding an open issue successfully', () => {
+        const octokit = new Octokit();
+        const github = new GithubUtils(octokit);
+        octokit.issues.listForRepo = jest.fn().mockResolvedValue({data: [issue]});
+        octokit.git.getRef = jest.fn().mockResolvedValue({data: []});
```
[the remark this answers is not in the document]
AndrewGable: Good catch!

lib/GithubUtils.js
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Takes in a GitHub issue object and returns the data we want
+     *
+     * @param {Object} issue
+     * @returns {Promise}
+     */
+    getStagingDeployCashData(issue) {
+        try {
+            const versionRegex = new RegExp('([0-9]+).([0-9]+).([0-9]+)(?:-([0-9]+))?', 'g');
```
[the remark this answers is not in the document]
roryabraham: I was wrong about needing to escape these characters, since you're using `new RegExp` instead of a regex literal.

roryabraham (approved):
LGTM 👍

PR #342, opened 2021-03-03
Remove botify from CLA

AndrewGable:
We no longer need to whitelist @botify as it does not commit or merge pull requests in this repo anymore, so I've removed it to make sure we are compliant!


sketchydroide (approved):
LGTM

PR #344, opened 2021-03-03
Updated dependencies to use main lodash package

nickmurray47:
### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/155566

Only affects the PubSub and combobox elements from JS-Libs, which appear in the Web-E, Web-S directory.

# Tests / Web QA
1. In project root, run `npm i` and then `npm link` to manually create the link
2. Have `grunt watch` running
3. Run `npm link expensify-common` in Web-E and Web-S project roots.
4. Test around on the site, and check the console.log to make sure nothing has broken.


nickmurray47:

nickmurray47:
Quick hold while I work out this error message `Cannot find module 'babelify' from '/Users/nicholasmurray/Expensidev/Web-Expensify/node_modules/expensify-common'`

nickmurray47:
Tried removing the `node_modules` in `Web-E` and `expensify-common` before going through the tests here, but still no dice.

thienlnam:
Travis tests are passing, is this good now or are you still running into that error message?

nickmurray47:
Still running into that error message, and the same behavior occurs when I checkout master.

nickmurray47:
Ok, figured out that we need to add `babelify` back to devDependencies to get this to work. Seems like it was removed for kinda unclear [reasons](https://github.com/Expensify/expensify-common/pull/290/files?file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.sh#r521096465) when this file was first ported over.

nickmurray47:
Off hold @thienlnam!

package.json
```diff
@@ -38,6 +37,7 @@
     "eslint-plugin-jest": "^24.1.5",
     "grunt": "*",
     "grunt-chokidar": "1.0.2",
+    "babelify": "10.0.0",
```
nickmurray47, line 40: cc @tgolen since you have more context than me. It seems like we removed this `devDependency` since it was [deprecated ](https://github.com/Expensify/expensify-common/pull/290/files?file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.sh#r521544886), but its removal is preventing us from properly running `npm link` for Web-E and Web-S, which significantly hinders development when testing out expensify-common changes.

I just wanted to make sure that this wouldn't introduce any breaking changes (and even if it did, it would just be on dev).

tgolen: It _probably_ won't break anything to put this back, but I'm curious what exactly the problem is with `npm link`. Is it throwing an error of some kind?

thienlnam (approved):
All yours! @tgolen

package.json
tgolen: Ah, thanks! Sorry, I missed it above. I guess there isn't much we can do about it. I'm not sure why it's using babelify from that repo, but since this is just a dev dependency, I don't think it's a big deal.

tgolen (approved):

PR #345, opened 2021-03-04
Add GithubUtils and generateVersionComparisonURL

SombreroElGringo:

I added the class GithubUtils and its method generateVersionComparisonURL.
I installed the following missing packages 'semver' and '@actions/core'.

PS: Also for the tests I needed to use async/await else I got some timeout. Also I tried different workaround for example expect(Promise).resolves.toBe(url) but it was not working correctly. At least with async/await we are sure the tests are working correctly.

### Fixed Issues
$ https://github.com/Expensify/Expensify.cash/issues/1577

# Tests

For repository with the following tags 2.2.2, 2.2.1, 2.1.0-1, 2.1.0, 2.0.0-1, 2.0.0, 1.2.2-2, 1.2.2-1, 1.2.2, 1.2.1, 1.2.0, 1.1.0, 1.0.4, 1.0.3, 1.0.2, 1.0.1, 1.0.0 and 0.0.1

MAJOR comparison
-- if we pass the tag 2.2.2 as param it should return https://<ETC>/1.2.2...2.2.2
-- if we pass the tag 1.2.0 as param it should return https://<ETC>/0.0.1...1.2.0

MINOR comparison
-- if we pass the tag 2.2.2 as param it should return https://<ETC>/2.1.0...2.2.2
-- if we pass the tag 1.1.0 as param it should return https://<ETC>/1.0.4...1.1.0

PATCH comparison
-- if we pass the tag 2.2.2 as param it should return https://<ETC>/2.2.1...2.2.2
-- if we pass the tag 1.0.0 as param it should return https://<ETC>/0.0.1...1.0.0

BUILD comparison
-- if we pass the tag 2.1.0-1 as param it should return https://<ETC>/2.1.0...2.1.0-1
-- if we pass the tag 1.2.2-2 as param it should return https://<ETC>/1.2.2...1.2.2-2

# QA

If the `generateVersionComparisonURL` return the good url following the parameters passed.

### DIFF
--- a/__tests__/GithubUtils-test.js
+++ b/__tests__/GithubUtils-test.js
@@ -42,39 +42,121 @@ const expectedResponse = {
     url: 'https://api.github.com/repos/Andrew-Test-Org/Public-Test-Repo/issues/29'
 };
 
-describe('GithubUtils.getStagingDeployCash', () => {
-    test('Test finding an open issue successfully', () => {
-        const octokit = new Octokit();
-        const github = new GithubUtils(octokit);
-        octokit.issues.listForRepo = jest.fn().mockResolvedValue({data: [issue]});
-        return github.getStagingDeployCash().then(data => expect(data).toStrictEqual(expectedResponse));
+describe('GithubUtils', () => {
+    describe('getStagingDeployCash', () => {
+        test('Test finding an open issue successfully', () => {
+            const octokit = new Octokit();
+            const github = new GithubUtils(octokit);
+            octokit.issues.listForRepo = jest.fn().mockResolvedValue({data: [issue]});
+            return github.getStagingDeployCash().then(data => expect(data).toStrictEqual(expectedResponse));
+        });
+
+        test('Test finding an open issue without a body', () => {
+            const octokit = new Octokit();
+            const github = new GithubUtils(octokit);
+
+            const noBodyIssue = issue;
+            noBodyIssue.body = '';
+
+            octokit.issues.listForRepo = jest.fn().mockResolvedValue({data: [noBodyIssue]});
+            return github.getStagingDeployCash()
+                .catch(e => expect(e).toEqual(new Error('Unable to find StagingDeployCash issue with correct data.')));
+        });
+
+        test('Test finding more than one issue', () => {
+            const octokit = new Octokit();
+            const github = new GithubUtils(octokit);
+            octokit.issues.listForRepo = jest.fn().mockResolvedValue({data: [{a: 1}, {b: 2}]});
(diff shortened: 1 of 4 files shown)


SombreroElGringo:

roryabraham:
recheck

roryabraham:
@SombreroElGringo Looks like you've got some merge conflicts to resolve.

lib/GithubUtils.js
```diff
@@ -60,4 +65,59 @@ export default class GithubUtils {
...
+
+                    if (semverLevel === 'BUILD'
+                        && semverSatisfies(repoTag, `<=${tagSemver.major}.${tagSemver.minor}.${tagSemver.patch}`)
+                    ) {
+                        resolve(getComparisonURL(repoTag, tagSemver));
+                        return true;
+                    }
+                    return false;
+                }))
+                .catch(githubError => reject(core.setFailed(githubError)));
```
roryabraham, line 120: I think this code should be unaware of the Github Actions it's meant to run in. So instead of using `core.setFailed` I would just do:

```suggestion
                .catch(githubError => reject(githubError));
```

package.json
```diff
@@ -18,6 +18,7 @@
     "url": "git+ssh://git@github.com/Expensify/JS-Libs.git"
   },
   "dependencies": {
+    "@actions/core": "^1.2.6",
```
roryabraham, line 21: And after making the other change you shouldn't need this dependency anymore.

__tests__/GithubUtils-test.js
```diff
@@ -42,39 +42,121 @@ const expectedResponse = {
...
+                },
+            }),
+        }));
+
+        const octokit = mockGithub().getOctokit();
+        const githubUtils = new GithubUtils(octokit);
+
+        describe('MAJOR comparison', () => {
+            test('should return a comparison url between 2.2.2 and 1.2.2', async () => {
+                const result = await githubUtils
```
roryabraham, line 100: Sadly Expensify doesn't use `async/await` in any of our Javascript code. There is a good explanation of this in our [style guide](https://github.com/Expensify/Expensify.cash/blob/master/STYLE.md#esnext-are-we-allowed-to-use-insert-new-language-feature-why-or-why-not). So you'll need to convert these to promises. Sorry 🙃

roryabraham (changes requested):

roryabraham:
Just a couple minor suggestions + the `async/await` -> `promise` changes. All minor stuff!

roryabraham (approved):
LGTM 👍

PR #347, opened 2021-03-05
[No QA] Create StagingDeployCash utilities

roryabraham:
@AndrewGable will you please review this?

We'll also need to create utilities to update existing `StagingDeployCash` issues, but I think this PR is big enough as-is and we can create the update function in a follow-up.

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/155336

# Tests
Automated unit tests created to cover these changes.

# QA
No QA.


lib/GithubUtils.js
```diff
@@ -119,4 +129,120 @@ export default class GithubUtils {
...
+
+    /**
+     * Parse the issue number from a URL.
+     *
+     * @param {String} URL
+     * @returns {String}
+     * @throws {Error} If the URL is not a valid Github Issue.
+     */
+    static getIssueNumberFromURL(URL) {
+        const matches = URL.match(/https:\/\/github\.com\/.*\/.*\/issues\/([0-9]+).*/);
```
AndrewGable, line 227: NAB - Could you use `RegExp` here? I feel like it makes regex a little bit more legible without needing to escape everything.

__tests__/GithubUtils-test.js
```diff
@@ -80,6 +85,161 @@ describe('GithubUtils', () => {
...
+                repos: {
+                    listTags: jest.fn().mockResolvedValue({data: mockTags}),
+                },
+            }),
+        }));
+
+        const octokit = mockGithub().getOctokit();
+        const githubUtils = new GithubUtils(octokit);
+
+        // TODO: mock octokit w/ tag so comparison URL is consistent.
```
AndrewGable, line 172: Still a todo?

__tests__/GithubUtils-test.js
```diff
@@ -80,6 +85,161 @@ describe('GithubUtils', () => {
...
+        });
+
+        test('Test all resolved deploy blockers', () => {
+            githubUtils.generateStagingDeployCashBody(tag, basePRList, basePRList, baseDeployBlockerList, baseDeployBlockerList)
+                .then((issueBody) => {
+                    // eslint-disable-next-line max-len
+                    expect(issueBody).toBe(`${allVerifiedExpectedOutput}${deployBlockerHeader}${closedCheckbox} ${baseDeployBlockerList[0]}\r\n${closedCheckbox} ${baseDeployBlockerList[1]}\r\n`)
+                });
+        });
+    })
```
AndrewGable, line 241: Is there a way to test that the output of `createNewStagingDeployCash` equals the [body here](https://github.com/Expensify/expensify-common/pull/347/files#diff-27894ed198bf5b3cf8699ff675f2ad04f08c0cdc89623579c98431b8c5c91627R21)? Basically let's make sure they will play well together, not just in isolation.

lib/GithubUtils.js
```diff
@@ -119,4 +129,120 @@ export default class GithubUtils {
...
+     * @param {String} tag
+     * @param {Array} PRList
+     * @returns {Promise}
+     */
+    createNewStagingDeployCash(title, tag, PRList) {
+        return this.octokit.issues.create({
+            owner: GITHUB_OWNER,
+            repo: EXPENSIFY_ISSUE_REPO,
+            labels: 'StagingDeployCash',
+            assignee: 'applausebot',
```
AndrewGable, line 146: Nab - Should these be constants?

__tests__/GithubUtils-test.js
roryabraham: Nope, comment removed.

__tests__/GithubUtils-test.js
roryabraham: Added a test

lib/GithubUtils.js
roryabraham: Updated to use constants 👍

lib/GithubUtils.js
roryabraham: Okay, so this has been updated to use `RegExp`.

_However_, our understanding of how `RegExp` works compared to `/someRegexLiteral/`was slightly off. You _do_ need to escape regex characters such as `.`, except since you're using a regular string, the first forward-slash `\` that you would typically use to escape a character in a regex is "consumed" by the string.

That's because strings have their own set of special characters which need to be escaped, such as `\n`, and by default if it's not a special character then Javascript will just strip the `\` from the string. So in plain-old JS, `console.log('\.');` outputs `.`, _not_ `\.`.

So, in order to match a period literal using `RegExp`, you need to escape the forward slash first, like this: `RegExp('\\.')`. However, because `/` is not a special character in a string, you don't need to escape that. You only need to escape `/` using a regex literal because it's the character used to start/stop a regex literal.

roryabraham:
@AndrewGable Ready for re-review

AndrewGable (approved):

stitesExpensify (approved):
LGTM!

stitesExpensify:
Nice tests btw

PR #349, opened 2021-03-06
Implement getPullRequestsMergedBetween function

tugbadogan:
@roryabraham @arielgreen will you please review this?

Added GitUtils class and implemented getPullRequestsMergedBetween with unit tests.

### Fixed Issues
$ https://github.com/Expensify/Expensify.cash/issues/1576

# Tests
I added unit test for `getPullRequestNumberFromCommitMessage` function

I manually executed following command and got expected result.

https://github.com/tugbadogan/expensify-common/blob/37f21ea035930a80282822f3dea6cd25e502d70b/index.js#L61-L67

# QA

### DIFF
--- a/__tests__/GitUtils-test.js
+++ b/__tests__/GitUtils-test.js
@@ -1,0 +1,17 @@
+import GitUtils from '../lib/GitUtils';
+
+describe('GitUtils', () => {
+    describe('getPullRequestNumberFromCommitMessage', () => {
+        test('Test matching PR number from commit message', () => {
+            const gitUtils = new GitUtils();
+            const pullRequestNumber = gitUtils.getPullRequestNumberFromCommitMessage("Merge pull request #337 from Expensify/francoisUpdateQbdSyncManager");
+            expect(pullRequestNumber).toStrictEqual('337');
+        });
+
+        test('Test non-matching PR number from commit message', () => {
+            const gitUtils = new GitUtils();
+            const pullRequestNumber = gitUtils.getPullRequestNumberFromCommitMessage("Update QBD Sync Manager version");
+            expect(pullRequestNumber).toBeNull();
+        });
+    });
+});
--- a/__tests__/GitUtils-test.js
+++ b/__tests__/GitUtils-test.js
@@ -1,17 +1,57 @@
-import GitUtils from '../lib/GitUtils';
+const GitUtils = require('../lib/GitUtils');
+const childproc = require('child_process');
+jest.mock('child_process');
+jest.mock('util', () => ({ promisify: (fn) => fn }));
+const { exec } = childproc;
+
+const data = [
+    {
+        gitLog: `Start adding StagingDeployCash logic
+        Setting up bones
+        Merge pull request #337 from Expensify/francoisUpdateQbdSyncManager
+        Merge pull request #336 from Expensify/andrew-pr-cla
+        Update QBD Sync Manager version
+        Only run CLA on PR comments or events
+        Merge pull request #331 from Expensify/marcaaron-killMoment
+        Merge pull request #330 from Expensify/andrew-cla-update
+        Merge pull request #333 from Expensify/Rory-AddOnOffSwitchTooltip
+        Setup OnOffSwitch component with tooltips
+        Merge pull request #332 from Expensify/alex-mechler-patch-1
+        Return to old hash-based deploy instrcutions
+        Remove DEFAULT_START_DATE & DEFAULT_END_DATE altogether`,
+        result: ['337', '336', '331', '330', '333', '332'],
+    },
+    {
+        gitLog: `Merge pull request #1521 from parasharrajat/parasharrajat/pdf-render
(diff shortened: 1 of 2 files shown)


tugbadogan:

lib/GitUtils.jsx
```diff
@@ -0,0 +1,44 @@
+const { promisify } = require('util');
+const exec = promisify(require('child_process').exec);
+
+export default class GitUtils {
```
roryabraham, line 4: Does this work properly? It's really odd to mix the CJS module system with the ES6 module system. You're importing your dependencies using CJS (aka `require()` instead of `import`), and then exporting using ES6 (`export default xxx`  instead of `module.exports = xxx`).

I think this may have only worked because the code is transpiled by `ncc` and bundled with its dependencies before being executed in the Github Actions runner. Since this code is meant to be executed in a Node.js context, let's stick with CJS modules for this file.

So I think you'd just need to replace `export default` with `module.exports = `

```suggestion
module.exports = class GitUtils {
```

...but be sure to retest :) You'll probably have to change the way you import it into the unit tests. Either `import * as GitUtils from '../lib/GitUtils';` or `const GitUtils = require('../lib/GitUtils');` would work I think.

__tests__/GitUtils-test.js
```diff
@@ -0,0 +1,17 @@
+import GitUtils from '../lib/GitUtils';
+
+describe('GitUtils', () => {
```
roryabraham, line 3: It would be great to see some unit tests that mock the output of the `exec('git log ...');` so you can test how the matching works with the actual output of a `git log` command.

So you could actually run that command for a couple different repos in your terminal, and then copy the output and use it as a mocked response for the `exec` command. That way you can test `getPullRequestsMergedBetween` in addition to `getPullRequestNumberFromCommitMessage`.

lib/GitUtils.jsx
```diff
@@ -0,0 +1,44 @@
...
+     * @param {String} fromRef
+     * @param {String} toRef
+     * @returns {Promise}
+     */
+    getPullRequestsMergedBetween(fromRef, toRef) {
+        return new Promise((resolve, reject) => {
+            exec(`git log --format="%s" ${fromRef}...${toRef}`)
+                .then(({ stdout, stderr }) => {
+                    const commitMessages = stdout.split('\n');
+                    const pullRequestNumbers = commitMessages
```
roryabraham, line 18: I think you could also use a global regex and `matchAll` to avoid looping and having a second function like this. I think this single function might achieve the same effect:

```javascript
getPullRequestsMergedBetween(fromRef, toRef) {
    return exec(`git log --format="%s" ${fromRef}...${toRef}`)
        .then(({ stdout, stderr }) => (
            [...stdout.matchAll(new RegExp('Merge pull request #(\\d{1,6})', 'g')]
                .map(match => match[1])
         ));
}
```

Also, you also don't necessarily need to wrap what you have in a `new Promise`.

roryabraham (changes requested):

lib/GitUtils.jsx
tugbadogan: 👍 I changed it to use `module.export` and used `require()` everywhere for consistency.

tugbadogan: The unit test is working as expected.

```
tugba ~/Documents/GitHub/expensify-common (tugbadogan-getPullRequestsMergedBetween)$ npm test GitUtils-test

> expensify-common@1.0.0 test /Users/tugba/Documents/GitHub/expensify-common
> jest "GitUtils-test"

 PASS  __tests__/GitUtils-test.js
  GitUtils
    getPullRequestsMergedBetween
      ✓ Test getPullRequestsMergedBetween by mocking exec (3 ms)
      ✓ Test getPullRequestsMergedBetween by mocking exec
      ✓ Test getPullRequestsMergedBetween by mocking exec (1 ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        1.356 s
Ran all test suites matching /GitUtils-test/i.
```

__tests__/GitUtils-test.js
tugbadogan: Sure, I added test for getPullRequestsMergedBetween by mocking exec.

lib/GitUtils.jsx
tugbadogan: 👍

__tests__/GitUtils-test.js
```diff
@@ -0,0 +1,57 @@
...
+        gitLog: `Return to old hash-based deploy instrcutions
+        Remove DEFAULT_START_DATE & DEFAULT_END_DATE altogether
+        refactor`,
+        result: [],
+    },
+];
+
+describe('GitUtils', () => {
+    describe.each(data)(`getPullRequestsMergedBetween`, (exampleCase) => {
+        test('Test getPullRequestsMergedBetween by mocking exec', () => {
```
roryabraham, line 50: NAB not sure you need to say "by mocking exec" here :)

roryabraham (approved):
Okay, code looks good!

PR #362, opened 2021-04-15
Update stripHTML to actually strip tags

yuwenmemon:
cc @tgolen

The way we were using `htmlDecode` in place of `stripHTML` was wrong I believe. Strip means to remove. In any case, the implementation of `htmlDecode` was not actually stripping tags in the first place:
```
>> Str.stripHTML('<strong>hello</strong>')
"<strong>hello</strong>"
>> $('<textarea/>').html('<strong>hello</strong>').text();
"<strong>hello</strong>"
```

### Fixed Issues
None, but using to clean up code mentioned here: https://github.com/Expensify/Expensify.cash/pull/2320/files#r613669999

# Tests
- Added Unit tests

### DIFF
--- a/__tests__/Str-test.js
+++ b/__tests__/Str-test.js
@@ -42,3 +42,12 @@ describe('Str.isValidURL', () => {
         expect(Str.isValidURL('test')).toBeFalsy();
     });
 });
+
+describe('Str.stripHTML', () => {
+    it('Correctly identifies valid urls', () => {
+        expect(Str.stripHTML('<strong>hello</strong>')).toBe('hello');
+        expect(Str.stripHTML('<img onerror=\'alert("could run arbitrary JS here")\' src=bogus>')).toBe('');
+        expect(Str.stripHTML('hello')).toBe('hello');
+        expect(Str.stripHTML(0)).toBe('');
+    });
+});
--- a/__tests__/Str-test.js
+++ b/__tests__/Str-test.js
@@ -44,7 +44,7 @@ describe('Str.isValidURL', () => {
 });
 
 describe('Str.stripHTML', () => {
-    it('Correctly identifies valid urls', () => {
+    it('Correctly strips HTML/XML tags', () => {
         expect(Str.stripHTML('<strong>hello</strong>')).toBe('hello');
         expect(Str.stripHTML('<img onerror=\'alert("could run arbitrary JS here")\' src=bogus>')).toBe('');
         expect(Str.stripHTML('hello')).toBe('hello');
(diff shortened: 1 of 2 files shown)


__tests__/Str-test.js
```diff
@@ -42,3 +42,12 @@ describe('Str.isValidURL', () => {
         expect(Str.isValidURL('test')).toBeFalsy();
     });
 });
+
+describe('Str.stripHTML', () => {
+    it('Correctly identifies valid urls', () => {
```
roryabraham, line 47: I think you'll want to rename this test

roryabraham (commented):
Did you look into how this function is used in Web-Expensify? I'm worried this is going to break stuff 😅

deetergp (commented):
Other than the copypasta test description, this looks good to me.

tgolen (changes requested):
Yeah, I agree that I am a bit worried that this could break other repos. Maybe we should add a new method instead and mark that one as deprecated? But first, let's check and see how often that method is used and where.

Would you mind cleaning up all the lint warnings? It looks like most of them are just lines that are too long.

roryabraham:
> Would you mind cleaning up all the lint warnings? It looks like most of them are just lines that are too long

These warnings have been here a while, but because they're all regex literals the lines can't really be broken up. Probably the best thing to do would be to just suppress them all I guess. (tempted to suggest we disable the max-length rule at the file level for that file).

tgolen:
Yep, I'm fine with suppressing some of them for things like regex
literals, though even regexes can be multi-line if you use the `x` (I
think) flag which ignores whitespace in the pattern. It looks like some
were just on comments though

yuwenmemon:
> But first, let's check and see how often that method is used and where.

Yeah, it's only used in two spots in Web-E. That's the only repo in which it's currently used:
```
yuwen@MemonbookProExpensify ~/Expensidev/Web-Expensify (yuwen-ghContributions) Web-Expensify % git grep stripHTML
site/app/dialogs/addCSV/addCSVView.jsx:                trans[i] = Str.stripHTML(row[layout[i]]);
site/app/dialogs/uploadCompanyCard/uploadCompanyCardView.jsx:                trans[fieldName] = Str.stripHTML(this.activeCSVObj[csvRowIndex][Number(fieldIndex)]).trim();
```

Checked out both those spots. They both meant to remove tags rather than escape them.

yuwenmemon:

yuwenmemon:
Updated!

tgolen (approved):
Thanks for looking into that.

roryabraham (approved):
:nooice:

PR #368, opened 2021-05-12
Add Inbox_Start_ScreenShare API command

francoisl:
### Fixed Issues
Needed for EW Screen Sharing project.

### Tests
The API command is still WIP in Web-E, this won't be used for now.
Will be tested with the rest of https://github.com/Expensify/Expensify/issues/163733.

### DIFF
--- a/lib/API.jsx
+++ b/lib/API.jsx
@@ -328,6 +328,20 @@ export default function API(network, args) {
         },
 
         /**
+         * Initiate a screen share request with a user.
+         *
+         * @param {object} parameters
+         * @param {Number} parameters.accountID
+         * @param {string} parameters.CallSid
+         * @returns {APIDeferred}
+         */
+        Inbox_Start_ScreenShare(parameters) {
+            const commandName = 'Inbox_Start_ScreenShare';
+            requireParameters(['accountID', 'CallSid'], parameters, commandName);
+            return performPOSTRequest(commandName, parameters);
+        },
+
+        /**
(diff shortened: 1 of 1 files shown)


lib/API.jsx
```diff
@@ -327,6 +327,20 @@ export default function API(network, args) {
             return performPOSTRequest(commandName, parameters);
         },
 
+        /**
+         * Initiate a screen share request with a user.
+         *
+         * @param {object} parameters
```
Beamanator, line 333: I believe we should make this upper-case

```suggestion
         * @param {Object} parameters
```

lib/API.jsx
```diff
@@ -327,6 +327,20 @@ export default function API(network, args) {
...
+         * @param {Number} parameters.accountID
+         * @param {string} parameters.CallSid
```
Beamanator, line 335: Also should be upper-case

```suggestion
         * @param {String} parameters.CallSid
```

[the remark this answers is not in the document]
Beamanator: Makes sense to me! I think it could be useful to add a comment explaining what the `CallSid` is, I don't think it's very self-explanatory, unless it's just a "call ID" (with an extra "s" thrown in) haha

Beamanator (changes requested):
Just a few tiny changes requested, then LGTM 👍

francoisl:
Updated

Beamanator (approved):
#LGTM! 🚀

PR #381, opened 2021-05-29
feat: HTML to markdown parser

pranshuchittora:
@roryabraham @jasperhuangg @marcaaron  will you please review this?

Context: https://github.com/Expensify/Expensify.cash/issues/2847

# Tests
**What unit/integration tests cover your change? What autoQA tests cover your change?**
Yes I have added tests

**What tests did you perform that validates your changed worked?**
Yes

# QA
**What does QA need to do to validate your changes?**
Try various HTML strings with `<br>` tags in various combinations

HTML -> MD
1. `Hello<br/>There` -> `Hello\nThere`
2. `Hello<br></br>There` -> `Hello\nThere`

Tests -> https://github.com/Expensify/expensify-common/pull/381/files#diff-7031bf36a0d060dfeb1742c45b8fa6853fb5790ef4da366b0fb96ceaafd342ea

**What areas to they need to test for regressions?**
Parser

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -157,6 +157,23 @@ export default class ExpensiMark {
                 replacement: '<br>',
             },
         ];
+
+        /**
+         * The list of regex replacements to do on a HTML comment for converting it to markdown.
+         *
+         * @type {Object[]}
+         */
+        this.htmlToMarkdownRules = [
+            {
+                name: 'newline',
+
+                // Replaces open and closing <br><br/> tags with a single <br/>
+                pre: inputString => inputString.replace('<br><br/>', '<br/>'),
+                regex: /<br\s*[\/]?>/gi,
+                replacement: '\n'
+            },
+
+        ];
     }
 
     /**
@@ -249,5 +266,24 @@ export default class ExpensiMark {
         if (startIndex < textToCheck.length) replacedText = replacedText.concat(textToCheck.substr(startIndex));
 
         return replacedText;
+    }
+
+    /**
+     * Replaces html with elements
+     *
+     * @param {String} htmlString
+     *
+     * @returns {String}
+     */
+    htmlToMarkdown(htmlString) {
+        let generatedMarkdown = htmlString;
+        this.htmlToMarkdownRules.forEach((rule) => {
+            // Pre-processes input HTML before applying regex
+            if (rule.pre) {
+                generatedMarkdown = rule.pre(generatedMarkdown);
+            }
+            generatedMarkdown = generatedMarkdown.replace(rule.regex, rule.replacement);
+        });
+        return generatedMarkdown;
     }
 }
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -169,7 +169,7 @@ export default class ExpensiMark {
 
                 // Replaces open and closing <br><br/> tags with a single <br/>
                 pre: inputString => inputString.replace('<br><br/>', '<br/>'),
-                regex: /<br\s*[\/]?>/gi,
+                regex: /<br\s*[\\/]?>/gi,
                 replacement: '\n'
             },
 
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -168,11 +168,10 @@ export default class ExpensiMark {
                 name: 'newline',
 
                 // Replaces open and closing <br><br/> tags with a single <br/>
-                pre: inputString => inputString.replace('<br><br/>', '<br/>'),
-                regex: /<br\s*[\\/]?>/gi,
+                pre: inputString => inputString.replace('<br></br>', '<br/>').replace('<br><br/>', '<br/>'),
+                regex: /<br\s*[/]?>/gi,
                 replacement: '\n'
             },
-
         ];
     }
 
@@ -269,7 +268,7 @@ export default class ExpensiMark {
     }
 
     /**
-     * Replaces html with elements
+     * Replaces HTML with markdown
      *
      * @param {String} htmlString
      *
--- a/__tests__/ExpensiMark-test.js
+++ b/__tests__/ExpensiMark-test.js
@@ -359,3 +359,9 @@ test('Test HTML string with seperate closing tags (<br><br/>) to markdown ', () 
 
     expect(parser.htmlToMarkdown(testString)).toBe(resultString);
 });
+
+test('Test HTML string with seperate closing tags (<br></br>) to markdown ', () => {
+    const testString = 'Hello<br>World,<br></br>Welcome<br/>To<br/>Expensify';
+    const resultString = 'Hello\nWorld,\nWelcome\nTo\nExpensify';
+    expect(parser.htmlToMarkdown(testString)).toBe(resultString);
+});
--- a/__tests__/ExpensiMark-test.js
+++ b/__tests__/ExpensiMark-test.js
@@ -338,3 +338,24 @@ test('Test markdown and url links with inconsistent starting and closing parens'
 
     expect(parser.replace(testString)).toBe(resultString);
 });
+
+test('Test HTML string with <br/> tags to markdown ', () => {
+    const testString = 'Hello<br/>World,<br/>Welcome<br/>To<br/>Expensify';
+    const resultString = 'Hello\nWorld,\nWelcome\nTo\nExpensify';
+
+    expect(parser.htmlToMarkdown(testString)).toBe(resultString);
+});
+
+test('Test HTML string with inconsistent <br/> closing tags to markdown ', () => {
+    const testString = 'Hello<br>World,<br/>Welcome<br>To<br/>Expensify';
+    const resultString = 'Hello\nWorld,\nWelcome\nTo\nExpensify';
+
+    expect(parser.htmlToMarkdown(testString)).toBe(resultString);
+});
+
+test('Test HTML string with seperate closing tags (<br><br/>) to markdown ', () => {
+    const testString = 'Hello<br>World,<br><br/>Welcome<br/>To<br/>Expensify';
+    const resultString = 'Hello\nWorld,\nWelcome\nTo\nExpensify';
+
+    expect(parser.htmlToMarkdown(testString)).toBe(resultString);
+});


pranshuchittora:

roryabraham:
@pranshuchittora I've run the test check suite on this PR and found that there are lint errors that need to be resolved

pranshuchittora:
@roryabraham I have fixed the linting issues :)

lib/ExpensiMark.js
```diff
@@ -157,6 +157,23 @@ export default class ExpensiMark {
...
+        this.htmlToMarkdownRules = [
+            {
+                name: 'newline',
+
+                // Replaces open and closing <br><br/> tags with a single <br/>
+                pre: inputString => inputString.replace('<br><br/>', '<br/>'),
+                regex: /<br\s*[\\/]?>/gi,
+                replacement: '\n'
+            },
+
```
Luke9389, line 175: Please remove this extra line of whitespace.

lib/ExpensiMark.js
```diff
@@ -157,6 +157,23 @@ export default class ExpensiMark {
...
+         *
+         * @type {Object[]}
+         */
+        this.htmlToMarkdownRules = [
+            {
+                name: 'newline',
+
+                // Replaces open and closing <br><br/> tags with a single <br/>
+                pre: inputString => inputString.replace('<br><br/>', '<br/>'),
+                regex: /<br\s*[\\/]?>/gi,
```
Luke9389, line 172: I have a few questions regarding our regex pattern: 
 - Why are we attempting to detect `<br\>`? 
 - What is the purpose of using the case insensitive flag?

lib/ExpensiMark.js
```diff
@@ -250,4 +267,23 @@ export default class ExpensiMark {
 
         return replacedText;
     }
+
+    /**
+     * Replaces html with elements
```
Luke9389, line 272: Can we change this to `Replaces HTML with markdown`?

lib/ExpensiMark.js
```diff
@@ -250,4 +267,23 @@ export default class ExpensiMark {
...
+     *
+     * @param {String} htmlString
+     *
+     * @returns {String}
+     */
+    htmlToMarkdown(htmlString) {
+        let generatedMarkdown = htmlString;
+        this.htmlToMarkdownRules.forEach((rule) => {
+            // Pre-processes input HTML before applying regex
```
Luke9389, line 281: Please add an empty line before this comment (like the one above).

Luke9389 (changes requested):
Hello @pranshuchittora. Nice work so far. I've got a few small comments and one larger one that I'll post below.

Luke9389:
Do we want to match breaking space tags that are inside of code blocks? 

If you sent a message like the following, what would we want our code to do? 
"Hey fellow engineer, I noticed we are missing a `<br/>` closing tag on line X." 

Currently, I think we'd end up matching and replacing that tag, which is not desired behavior. Perhaps we should establish a method for preventing matches inside of code blocks. Thoughts?

Luke9389:
As a side note, I think you can put the changes from this PR here as well. I like the organization of having them separate, but I think it's best for our workflow to just have one PR to track.

lib/ExpensiMark.js
pranshuchittora: Hi, 
1. we can remove `\` regex
2. This is because HTML is a case insensitive language 

I am thinking of adding `<br></br>` as well in `pre`

pranshuchittora:
> "Hey fellow engineer, I noticed we are missing a <br/> closing tag on line X."

As this happens on runtime (conversion), therefore this becomes irrelevant. 
My approach / philosophy is inspired by the way HTML works instead of breaking it assumes and fixes the tree on runtime

lib/ExpensiMark.js
Luke9389: Ok great! Thx. This should work nicely, idk if the `pre` is necessary.

pranshuchittora: Initially having a regex worked, but later when I was writing some tests and figuring out edge cases. It became necessary to add pre.

Usually, some sanitisers work by appending a closing tag for the uncloses tags `<br>`. That is the reason behind adding pre. Or you can say my addiction to defensive programming 😅

pranshuchittora: I am also accounting for both HTML style and react style of closing

lib/ExpensiMark.js
pranshuchittora: Screenshot from 2021-06-05 20-53-10

Luke9389: Ah, yea, that's due to the whitespace on that line. If you backspace a bunch of times it'll get rid of that linting error.

Luke9389:
> As this happens on runtime (conversion), therefore this becomes irrelevant. My approach / philosophy is inspired by the way HTML works instead of breaking it assumes and fixes the tree on runtime

Interesting. Would you mind elaborating a bit further, for my sake? Also, is this something we can test?

pranshuchittora:
> > As this happens on runtime (conversion), therefore this becomes irrelevant. My approach / philosophy is inspired by the way HTML works instead of breaking it assumes and fixes the tree on runtime
> 
> Interesting. Would you mind elaborating a bit further, for my sake? Also, is this something we can test?

Sure, what I understand is that we should show some warning for unclosed `<br>` in dev mode. This kinda becomes as messages are going to be generated by the users on runtime. 

The only way this becomes relevant is when input to HTML conversion happens. Instead of handling it here, I recommend writing robust test cases for the HTML conversion parser.

Luke9389:
> Sure, what I understand is that we should show some warning for unclosed `<br>` in dev mode.
 
Ok it's possible we're talking about two different things right now. My main concern is that if someone puts a br tag inside backticks, that we won't make a newline. I'm not sure what you're referring to with showing a warning.

pranshuchittora:
> > Sure, what I understand is that we should show some warning for unclosed `<br>` in dev mode.
> 
> Ok it's possible we're talking about two different things right now. My main concern is that if someone puts a br tag inside backticks, that we won't make a newline. I'm not sure what you're referring to with showing a warning.

Yeah, I agree that's an edge case. I try finding out how other parsers are doing this unfortunately one of the famous parser has buggy behaviour with this as well.

http://domchristie.github.io/turndown/

Paste this HTML 👇🏼 
```
<code>
Hello There <br/>
</code>
```

jasperhuangg (approved):

Luke9389:
Couldn't we just use regex to accomplish this? I think negative lookahead will allow us to avoid matching a breaking tag wrapped in backticks. Something like this: <pre><br\s*[/]?>(?!`)</pre>

That way we won't match `<br/>`

pranshuchittora:
> I think negative lookahead will allow us to avoid matching a breaking tag wrapped in backticks

@Luke9389 backticks are from the markdown world.
Backticks are converted to `<code>` in HTML. The input to this will look like this
```html
<code>
Hello There <br/>
</code>

```

lib/ExpensiMark.js
Luke9389: Did this get resolved?

pranshuchittora:
Hi, @marcaaron can you help merging this

marcaaron:
Sorry, @pranshuchittora please give the assigned reviewers time to handle this and if > 3 days ask in the Slack channel for help

Luke9389:
Hey @pranshuchittora, sorry for the late response here. I've been on vacation the past week. 

Ok, I see what you're saying about the backticks, but we aren't out of the weeds yet.

First, I want to be sure we're on the same page. When someone sends a message with this, `<br/>`, they're talking about a br tag, and so we shouldn't parse that to a new line. That gets sent, turned into HTML (< code >< br/ >< /code >), and now when we are parsing that HTML back to markdown, we want to be sure we don't create a new line (we want it to be turned into `<br/>`). 

So in my view, we should have some way of preventing the content inside of code blocks from being parsed. Maybe that should be saved for a different PR? What do you think @pranshuchittora?

jasperhuangg (changes requested):
I agree with Luke's comments about ensuring that any code inside of code blocks isn't parsed.

pranshuchittora:
> Hey @pranshuchittora, sorry for the late response here. I've been on vacation the past week.
> 
> Ok, I see what you're saying about the backticks, but we aren't out of the weeds yet.
> 
> First, I want to be sure we're on the same page. When someone sends a message with this, `<br/>`, they're talking about a br tag, and so we shouldn't parse that to a new line. That gets sent, turned into HTML (< code >< br/ >< /code >), and now when we are parsing that HTML back to markdown, we want to be sure we don't create a new line (we want it to be turned into `<br/>`).
> 
> So in my view, we should have some way of preventing the content inside of code blocks from being parsed. Maybe that should be saved for a different PR? What do you think @pranshuchittora?

Yup we are on the same page 👍🏼 
Solving this is very tricky as we can’t find `<code>` directly as it can have some attribute like `<code some-attribute="xyz" >`. I checked few parsers out there as discussed in this comment https://github.com/Expensify/expensify-common/pull/381#issuecomment-856948421

IMO let’s solve that with another issue and PR

Luke9389:
Yea, I think solving that in another PR is reasonable. It'll be its own issue with its own solution. I'm ready to merge as soon as this gets resolved: https://github.com/Expensify/expensify-common/pull/381/files#r644268511

lib/ExpensiMark.js
pranshuchittora: I tried but it's not working 

Screenshot from 2021-06-15 22-53-26

pranshuchittora:
> Yea, I think solving that in another PR is reasonable. It'll be its own issue with its own solution. I'm ready to merge as soon as this gets resolved: https://github.com/Expensify/expensify-common/pull/381/files#r644268511

https://github.com/Expensify/expensify-common/pull/381/files#r652000032

lib/ExpensiMark.js
Luke9389: Hmm, OK. Not a big deal I guess. Thanks for trying

Luke9389 (approved):
One more thing, can you add specific tests (numbered and detailed) for what Web QA should do to test this? Try and make it as easy for them as possible to understand what they need to do. Right now it's too vague.

jasperhuangg (approved):

PR #396, opened 2021-07-20
fixed url parsing for escaped chars

parasharrajat:
Anyone, will you please review this?

We escape the input string before apply Rules from Expensimark but we forgot to check for the Escaped chars in the URL regex. I have updated that.
Apart from it, Added `:` to the regex so that it can be matched as part of the URL. Google Map URLs contain this.

### Fixed Issues
$ https://github.com/Expensify/App/issues/4105

# Tests
Unit test
I tried many URLs for matching and all seem to pass 

# QA
  URLs should be parsed in E.cash.
E.cash messaging


parasharrajat:
@Jag96 Requesting review. Thanks.

tgolen (approved):

PR #402, opened 2021-07-28
Quote support for HTML => MD Conversion 

parasharrajat:
TAG_REVIEWER will you please review this?

Added support for HTML quote parsing in HTML to MD conversion.
1. It first grabs the whole string between `<blockquote>` tags.
2. we divide each new line in the text with `> line text`
3. if line Text is `\n` we skip this char to remove the extra line and conversion would be `> `.

 
### Fixed Issues
$ https://github.com/Expensify/App/issues/4188

# Tests
   I have added Unit tests that cover cases where Quotes can also contain other HTML tags or newlines.

    I used a couple of quoted texts and paste them into the Composer in the E.cash to test my solution. 

# QA
  Try to paste copied quotes from other websites to the E.cash.

1. What areas do they need to test for regressions?
   Messaging on E.cash.

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -214,6 +214,14 @@ export default class ExpensiMark {
                 regex: /<(del)(?:"[^"]*"|'[^']*'|[^'">])*>(.*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: '~$2~'
             },
+            {
+                name: 'quote',
+                regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
+                replacement: (match, g1, g2) => {
+                    const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
+                    return `\n${resultString}\n`;
+                }
+            },
         ];
     }
 
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -218,7 +218,7 @@ export default class ExpensiMark {
                 name: 'quote',
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
-                    const resultString = g2.trim().split('\n').map(m => `> ${m || ''}`).join('\n');
+                    const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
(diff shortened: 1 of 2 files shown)


roryabraham:
Grabbing this because @Luke9389 is OOO

roryabraham:
@parasharrajat Please fill out the PR template

parasharrajat:
@roryabraham  Updated.

lib/ExpensiMark.js
```diff
@@ -214,6 +214,14 @@ export default class ExpensiMark {
                 regex: /<(del)(?:"[^"]*"|'[^']*'|[^'">])*>(.*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: '~$2~'
             },
+            {
+                name: 'quote',
+                regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
```
roryabraham, line 219: Can you explain what's going on with this portion of the regex?

```
(?:"[^"]*"|'[^']*'|[^'">])
```

Is this just to allow for attributes in the opening HTML tag? If so, can you include some such examples in the unit test?

---

I also have a question about the second capturing group:

```
([\s\S]*?)
```

Is the `?` just so that it doesn't go beyond the next opening tag? If so, do you think this is maybe simpler to understand?

```
([^<]*)
```

lib/ExpensiMark.js
```diff
@@ -214,6 +214,14 @@ export default class ExpensiMark {
...
+                replacement: (match, g1, g2) => {
+                    const resultString = g2.trim().split('\n').map(m => `> ${m || ''}`).join('\n');
```
roryabraham, line 221: Also can we safely simplify this like so:

```suggestion
                    const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
```

roryabraham: Asking earnestly, but I don't think `.split` would give us falsey array elements, right?

lib/ExpensiMark.js
parasharrajat: > (?:"[^"]*"|'[^']*'|[^'">])

Yes, attributes. I will add tests

> ([\s\S]*?)

 here we are matching everything. *? is non-greedy so it will end the match at first </blockquote>.

> ([^<]*)

Can't use it it does not match other HTML tags. Blockquote can have other tags e.g. `b,a del, em` etc.

lib/ExpensiMark.js
parasharrajat: Correct.

parasharrajat:
Updated.

roryabraham (approved):
Looking good

PR #405, opened 2021-08-01
Code fence HTML to markdown conversion

tugbadogan:
TAG_REVIEWER will you please review this?

Added regex rule for code fence HTML to markdown conversion.

### Fixed Issues
$ https://github.com/Expensify/App/issues/4184

# Tests
I added unit test for this regex rule.
I run E.cash with updated Expensify-common package from this PR.
Copy/pasted sample code pieces from Slack to E.cash.
Used edit feature on E.cash to make sure code fences are converted to markdown properly.

https://user-images.githubusercontent.com/3853003/127780267-88f46d9d-0389-45d9-9d9d-d81bcfab8bf0.mov

# QA
Copy/paste code blocks from Slack to E.cash
Edit code block messages and save again.
Make sure there is no change after edit.
Make sure other HTML markdown changes are not broken

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -196,7 +196,7 @@ export default class ExpensiMark {
                 pre: inputString => inputString.replace('<br></br>', '<br/>').replace('<br><br/>', '<br/>'),
 
                 // Include the immediately followed newline as `<br>\n` should be equal to one \n.
-                regex: /<br(?:"[^"]*"|'[^']*'|[^'"><])*>(?![^<]*(<\/pre>|<\/code>))\n?/gi,
+                regex: /<br(?:"[^"]*"|'[^']*'|[^'"><])*>\n?/gi,
                 replacement: '\n'
             },
             {
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -227,6 +227,11 @@ export default class ExpensiMark {
                 regex: /<(code)(?:"[^"]*"|'[^']*'|[^'">])*>(.*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: '`$2`'
             },
+            {
+                name: 'codeFence',
+                regex: /<(pre)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
+                replacement: '```\n$2\n```'
+            },
         ];
     }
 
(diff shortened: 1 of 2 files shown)


lib/ExpensiMark.js
```diff
@@ -196,7 +196,7 @@ export default class ExpensiMark {
                 pre: inputString => inputString.replace('<br></br>', '<br/>').replace('<br><br/>', '<br/>'),
 
                 // Include the immediately followed newline as `<br>\n` should be equal to one \n.
-                regex: /<br(?:"[^"]*"|'[^']*'|[^'"><])*>(?![^<]*(<\/pre>|<\/code>))\n?/gi,
```
tugbadogan, line 199: I had to delete this part to make sure new lines are properly replaced within `<pre>` tag. Otherwise, we're loosing the last new line in the code block.

Here's an example text where it wouldn't work correctly before this change: 
```
<pre>a<br />a<br />a</pre>
```

We're getting the following output string: 
``````
```
a
aa
```
``````

Screenshot 2021-08-01 at 18 31 41

**Review by MariaHCD — COMMENTED** [review]
Looks good!

While testing I noticed that new lines aren't being converted properly. Example:

Screen Shot 2021-08-03 at 4 11 40 PM
Screen Shot 2021-08-03 at 4 11 48 PM

**Comment by tugbadogan** [comment]
> Looks good!
> 
> While testing I noticed that new lines aren't being converted properly. Example:
> 
> Screen Shot 2021-08-03 at 4 11 40 PM
> Screen Shot 2021-08-03 at 4 11 48 PM

I tested this scenario by copying a code block with empty lines from Slack. I found out that Slack uses special `<span>` tag for empty lines. I modified newline HTML to markdown replacement to handle Slack's `<span>` tag for empty lines. It works now.

File: lib/ExpensiMark.js [hunk 2]
```diff
@@ -227,6 +230,11 @@ export default class ExpensiMark {
                 regex: /<(code)(?:"[^"]*"|'[^']*'|[^'">])*>(.*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: '`$2`'
             },
+            {
+                name: 'codeFence',
+                regex: /<(pre)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
```
parasharrajat, line 235: is it good to capture existing newlines before and after pre-tag boundaries?
Otherwise, we will end up adding extra line breaks.

MariaHCD: @parasharrajat Would you mind expanding on how we will end up with extra line breaks?

MariaHCD (approved):
Looks good to me so far and tests worked as expected.

Would appreciate @Jag96 and @parasharrajat's thoughts!

lib/ExpensiMark.js
parasharrajat: Let's say the code block has space before and After. In HTML, it looks like

```
<br />
<pre>
this is a pre tag with some conten
</pre>
<br/> 
```
Here the conversion will happen as follows based on the current Code.
=> BR tag
```
\n
<pre>
this is a pre tag with some conten
</pre>
\n
```
=> PRE Tag
```
\n
```
this is a pre tag with some content
```
\n
```

Here I think those newlines should not be present.

tugbadogan: I think if code block has empty lines before and after, we should keep those and replace them with new lines to have consistent markdown <-> HTML conversion. 

I tested with empty lines before code block and it works as expected (when you click Edit button and then Save button, final message should be exactly the same as first message).

parasharrajat: Ok.Great if that works. We can always look back and update it.

Jag96 (approved):
LGTM

PR #406, opened 2021-08-03
Links support to HTML => MD conversion

parasharrajat:
1. Firstly, it parses the Anchor tags.
2. Now based on found groups in the match. It first picks links that start with `mailto:` and then if text and URL are not the same. It would be converted to Markdown style links `[]()`.
3. At last, if step 2 is not a pass, then it would be a normal link.

I noticed that it's unsafe to unescape the HTML before matching tags. for example, if a tag is escaped then it means it was not HTML and thus I moved the `_.unscape` after all the parsing is done.

### Fixed Issues
$ Fixes https://github.com/Expensify/App/issues/4187

# Tests
I have a lot of Unit tests.
I used dummy html text to test the conversion.

# QA
  Pasting a link on NotDot should be converted to MD-style links.

1. What areas do they need to test for regressions?
  NewDot Message composing.

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -227,6 +227,17 @@ export default class ExpensiMark {
                 regex: /<(code)(?:"[^"]*"|'[^']*'|[^'">])*>(.*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: '`$2`'
             },
+            {
+                name: 'anchor',
+                regex: /<(a)[^><]*href\s*=\s*(['"])(.*?)\2(?:".*?"|'.*?'|[^'"><])*>(.*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
+                replacement: (match, g1, g2, g3, g4) => {
+                    const link = g3.startsWith('mailto:') ? g3.slice(7) : g3;
+                    if (link !== g4) {
+                        return `[${g4}](${link})`;
+                    }
+                    return link;
+                }
+            },
         ];
     }
 
@@ -330,8 +341,7 @@ export default class ExpensiMark {
      * @returns {String}
      */
     htmlToMarkdown(htmlString) {
-        let generatedMarkdown = _.unescape(htmlString);
-
+        let generatedMarkdown = htmlString;
         this.htmlToMarkdownRules.forEach((rule) => {
             // Pre-processes input HTML before applying regex
             if (rule.pre) {
@@ -339,7 +349,7 @@ export default class ExpensiMark {
             }
             generatedMarkdown = generatedMarkdown.replace(rule.regex, rule.replacement);
         });
-        return Str.stripHTML(generatedMarkdown);
+        return _.unescape(Str.stripHTML(generatedMarkdown));
     }
 
     /**
--- a/__tests__/ExpensiMark-Markdown-test.js
+++ b/__tests__/ExpensiMark-Markdown-test.js
@@ -118,12 +118,12 @@ test('Test HTML string with encoded entities', () => {
 
 test('Test HTML string with blockquote', () => {
     const testString = '<blockquote><p>This GH seems to assume that there will be something in the paste\nbuffer when you copy block-quoted text out of slack. But when I dump\nsome <em>lorem ipsum</em> into a blockquote in Slack, copy it to the\nbuffer, then dump it into terminal, there\'s nothing. And if I dump it </blockquote>'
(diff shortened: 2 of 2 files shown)


lib/ExpensiMark.js
```diff
@@ -330,16 +341,15 @@ export default class ExpensiMark {
      * @returns {String}
      */
     htmlToMarkdown(htmlString) {
-        let generatedMarkdown = _.unescape(htmlString);
-
+        let generatedMarkdown = htmlString;
```
MonilBhavsar, line 344: Won't this affect other conversions, as we would have regex for processing unescaped strings

parasharrajat: I didn't understand completely. Do you mean that other tags parsing in HTM=> MD conversion or All the conversions HTML=> MD, MD=> HTML?

MonilBhavsar: Yes other tags parsing in HTML->MD

parasharrajat: @MonilBhavsar I moved this down after the regex match based on two conditions.

1. First, If HTML contains an escaped string that can contain escaped tags, then that string is not parsed at HTML. it was escaped so that it shows as a raw string. Thus it does not have an effect on HTML parsing.
For example. If you can to show in the UI `<b>Rajat</b>` as a string in HTML. 
You can't write it as `<p> <b>Rajat</b> </p>`. It would be shows as <b>Rajat</b>. Instead you have to wirte `<p>&lt;b&gt;Rajat&lt;/b&gt;`.

2. In href, URLs, `"` could be used as &quot; to keep the HTML valid. But if unescape it before matching them it becomes `"` which breaks the URL parsing.
for example <a href="http://a-&quot;url">test</a>.
here url will be parsed as http://a- if we unescape ealier.

3. I added this _.unesacpe earlier to show the chars correctly. 
take the example from 1, `<p>&lt;b&gt;Rajat&lt;/b&gt;</p>`. will give as `&lt;b&gt;Rajat&lt;/b&gt;` after parsing but on screen you want to see `<b/>Rajat</b>`.

Hope it answers your question. Please let me know if I missed some cases. I will rethink that.

MonilBhavsar: Not sure If this is right, I follow that, 

1. If html contains escaped characters, it's a raw string and it won't be parsed here? Right now if I write `&lt;p&gt;&lt;b&gt;Monil&lt;/b&gt;&lt;/p&gt;` then I see `<p><b>Monil</b></p>`

2. Makes sense, that's why we are not unescaping it before. 

3. Sorry can't follow this.

My take, Let' say I type `<em>Monil</em>` then it should parse italics regex rule and show as `_Monil_`(_Monil_). And when I copy this, I get `&lt;em&gt;Monil&lt;/em&gt;`. So this is not working as expected right now, right?

parasharrajat: Sorry for the typos in the earlier message, 
1. If you paste `<em>Monil</em>`, you will get `_Monil_`.  But if you type this `<em>Monil</em>` its is taken as HTML , we don't support HTML markup in composer. It eould be saved as `&lt;em&gt;Monil&lt;/em&gt;`.

2. take the example from 1, if you pase HTML as `<p>&lt;b&gt;Rajat&lt;/b&gt;</p>`. Here I have used escaping to differentiate tages from text. But at the source location In UI `&lt;b&gt;Rajat&lt;/b&gt;` is shown as `<b>Rajat</b>` raw text.

Now if you parse `<p>&lt;b&gt;Rajat&lt;/b&gt;</p>`, it will give as `&lt;b&gt;Rajat&lt;/b&gt;` text in expensify. But as a user you don't want to see this in composer. A user would like to see it as  `<b/>Rajat</b>` But it does not mean that it is HTML. Only for this reason we have added unescaping.

@MonilBhavsar Where are copying from?

MonilBhavsar: Makes sense now. Thanks.
Messages with html tags will br parsed straight away by our regex rules.
Messages with escaped characters won't, we'll just unescape them and display it.

Right?
If I copy and paste - `<em>Monil</em>` and `<b>Monil</b>` from github and send message in App, they are sent as HTML raw code.
But If I copy that sent message and paste it again, I see converted forms like `_Monil_` and `_*Monil*_` in composer

parasharrajat: it does not matter what is shown in UI, the main thing is what is copied as HTML. Test what do you have `text/html` http://madebyevan.com/clipboard-test/ .

MonilBhavsar: Okay, works when type is `text/html`

MonilBhavsar (approved):

PR #411, opened 2021-08-25
fix: decode html entitites

parasharrajat:
@Beamanator  will you please review this?

### Fixed Issues
$ https://github.com/Expensify/App/issues/4759

# Tests
	The unit test contains HTML entities.
  I applied the fix on the NewDot and try pasting a few messages from google Docs which can contain HTML entities.

# QA
    Check that any HTML entity is being decoded correctly

    NewDot

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -356,7 +356,7 @@ export default class ExpensiMark {
             }
             generatedMarkdown = generatedMarkdown.replace(rule.regex, rule.replacement);
         });
-        return _.unescape(Str.stripHTML(generatedMarkdown));
+        return Str.htmlDecode(Str.stripHTML(generatedMarkdown));
     }
 
     /**


MonilBhavsar:
Looks good except Linter error

parasharrajat:
@MonilBhavsar Updated.

MonilBhavsar:
Also, I just saw the proposal was to extract characters from 3rd party library and parse them using regex. So was that plan updated?

parasharrajat:
I Just found that we are already using this lib in our bundle so I didn't have to do that

MonilBhavsar:
Yes, I see we are already using `html-entities`. Thanks for clarifying.

MonilBhavsar:
Lol, what am I seeing this. Github bug may be.

Screenshot 2021-08-25 at 1 49 13 PM

parasharrajat:
You mean the Lint warning?

It's not a GitHub bug, in our eslint config max-length rule is set to warning so lint test does not fail for this but a warning annotation is added.

MonilBhavsar:
No, Line 0 that shows parenthesis was removed

parasharrajat:
Yeah, It's a bug.

MonilBhavsar (approved):
@Beamanator, all yours

Beamanator (commented):
@parasharrajat you mention "The unit test contains HTML entities." - can you please mention which one you're talking about? Also if you're fixing something that's broken, shouldn't we update the test to make sure it doesn't break again? Please let me know if you feel that's unnecessary and if so, why

parasharrajat:
There are some urls that had the entities but yeah it is good to add a new test. I will add that. How does that sound?

Beamanator:
I think that's best 👍 Thanks so much dude

PR #423, opened 2021-10-20
Add Expensiworks_GetJob to lib/API.jsx

Luke9389:
### Part of:
https://github.com/Expensify/Expensify/issues/179284

### Tests
This will be tested in the PR to accompany this one on the linked issue.

### DIFF
--- a/lib/API.jsx
+++ b/lib/API.jsx
@@ -335,7 +335,6 @@ export default function API(network, args) {
          */
          Expensiworks_GetJob(parameters) {
             const commandName = 'Expensiworks_GetJob';
-            requireParameters(['jobName'], parameters, commandName);
             return performPOSTRequest(commandName, parameters);
         },
 


Luke9389:
Hmm, something's up with pullerbear. Looks like maybe it's auto assigned as a code owner when we commit to PRs that exist, but doesn't get auto-assigned when we create a PR that already has commits on the branch...

Luke9389:
This seems to be the way we are accessing the Expensiworks API commands from `inboxCall.jsx`. Here's an example from the same file: https://github.com/Expensify/Web-Expensify/blob/07b8481e478e228fd3b223b007ede2662a8d686e/site/app/modals/InboxCall.jsx#L36

Luke9389:
I have a 1:1 right now, but I'll post in eng chat after.

Luke9389:
OK now that I look at this again, you're totally right. We probably shouldn't expose an endpoint in an open source repo if we don't have to.

PR #424, opened 2021-11-04
Fix combobox searching

stitesExpensify:
We were adding non-matches to searches when `this.props.showDisabledOptionsInResults` was true. This was causing the search to not work for very large policies.

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/147788

# Tests/QA
1. Make this change in Web-Expensify 
    - i.e. open `./Web-Expensify/node_modules/expensify-common/lib/components/form/element/combobox.js`, apply the diff from this PR, and run `grunt` in `./Web-Expensify`
2. Turn on local js for prod
3. Supportal into ops@kruzeconsulting.com
4. Go to domains -> wellapp.com -> company cards
5. Click "Edit exports" on one of the users
6. Under QBO where it says "Default Card" search for whatever you want
7. Make sure the results actually match what you type

### DIFF
--- a/lib/components/form/element/combobox.js
+++ b/lib/components/form/element/combobox.js
@@ -768,7 +768,7 @@ class Combobox extends React.Component {
                     // Don't include the disabled options that match the regex unless we specified we want them to show.
                     // If we want them to show then add them whether they match the regex or not.
                     if ((!isDisabled && isMatch)
-                        || (isDisabled && this.props.showDisabledOptionsInResults)) {
+                        || (isDisabled && this.props.showDisabledOptionsInResults && isMatch)) {
                         matches.add(option);
                     }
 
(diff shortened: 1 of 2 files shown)


stitesExpensify:

stitesExpensify:
recheck

lib/components/form/element/combobox.js
```diff
@@ -767,8 +767,7 @@ class Combobox extends React.Component {
 
                     // Don't include the disabled options that match the regex unless we specified we want them to show.
                     // If we want them to show then add them whether they match the regex or not.
-                    if ((!isDisabled && isMatch)
-                        || (isDisabled && this.props.showDisabledOptionsInResults)) {
+                    if ((!isDisabled && isMatch) || (isDisabled && this.props.showDisabledOptionsInResults && isMatch)) {
```
stitesExpensify, line 770: The only difference besides style is that I added `&& isMatch` to the second condition because by the old logic, we would add every single non-match to matches lol

jasperhuangg: Isn't this just the same as `if (isDisabled && isMatch)`?

`if (A || A & B)` is always going to be true if `A` is true, so it simplifies to `if (A)`.

jasperhuangg (approved):
Tested and works great!

PR #431, opened 2022-01-06
Add a new constant for receipts URL

flodnv:
### Fixed Issues
For https://github.com/Expensify/Expensify/issues/71013

# Tests
See Associated PR

### DIFF
--- a/lib/CONST.jsx
+++ b/lib/CONST.jsx
@@ -348,6 +348,10 @@ export const CONST = {
             DEV: 'https://s3.amazonaws.com/dev-receipts-expensify-com/',
             PROD: 'https://s3.amazonaws.com/receipts.expensify.com/'
         },
+        RECEIPTS: {
+            DEV: 'https://www.expensify.com.dev/receipts/',
+            PROD: 'https://www.expensify.com/receipts/',
+        },
         CLOUDFRONT: 'https://d2k5nsl2zxldvw.cloudfront.net',
         CLOUDFRONT_IMG: 'https://d2k5nsl2zxldvw.cloudfront.net/images/',
         CLOUDFRONT_FILES: 'https://d2k5nsl2zxldvw.cloudfront.net/files/',


flodnv:

flodnv:
recheck

lib/CONST.jsx
```diff
@@ -348,6 +348,10 @@ export const CONST = {
             DEV: 'https://s3.amazonaws.com/dev-receipts-expensify-com/',
             PROD: 'https://s3.amazonaws.com/receipts.expensify.com/'
         },
+        RECEIPTS: {
+            DEV: 'https://www.expensify.com.dev/receipts/',
+            PROD: 'https://www.expensify.com/receipts/',
```
flodnv, line 353: I proposed something else (better?) but IDK if it's possible: https://expensify.slack.com/archives/C03TQ48KC/p1641495058018300?thread_ts=1641494910.018000&cid=C03TQ48KC

joelbettner: I'm not sure how the proposal you linked is different from what you are doing here...

joelbettner (approved):

iwiznia (approved):

lib/CONST.jsx
```diff
@@ -348,6 +348,10 @@ export const CONST = {
             DEV: 'https://s3.amazonaws.com/dev-receipts-expensify-com/',
             PROD: 'https://s3.amazonaws.com/receipts.expensify.com/'
         },
+        RECEIPTS: {
```
iwiznia, line 351: Can't comment above... is `BUCKET_RECEIPT` unused now?

flodnv: I think so, but I want to deploy the whole thing before removing it (specifically in case it needs to be rolled back) 

Issue here: https://github.com/Expensify/Expensify/issues/191720

lib/CONST.jsx
flodnv: I want it to make it dynamic dependent on an environment variable telling us whether it's staging or prod, but thinking more about this, I also thought that the calling code should do that and it's fine (like in my associated PR)

PR #438, opened 2022-04-05
Replace underscore.uniq by lodash.uniqBy for performance

aldo-expensify:
Underscore's `uniq` function is slow as **** in comparison with lodash's `uniqBy`. This performance issue was causing the browser to freeze when used on a big array (205141 elements in this case)

Doing tests locally, `underscore.uniq` just freezes, don't know how long it really takes, meanwhile, `lodash.uniqBy` returns in around 200 ms

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/204905

# Tests
1. Do the same changes this PR is doing in `node_modules/expensify-common/lib/components/form/element/combobox.js` (Web-E)
2. Use `grunt watch` to rebuild everything
3. Use this [SO](https://stackoverflow.com/c/expensify/questions/29) to use your local JS in production
4. Follow the QA Steps

# QA
1. Supportal into accounting@alayacare.com
5. Type something in this combobox to filter the results:
image
6. Should not freeze (it may be slow, but not freeze!)

### DIFF
--- a/lib/components/form/element/combobox.js
+++ b/lib/components/form/element/combobox.js
@@ -5,6 +5,7 @@ import cn from 'classnames';
 import _ from 'underscore';
 import get from 'lodash/get';
 import has from 'lodash/has';
+import uniqBy from 'lodash/uniqBy';
 import Str from '../../../str';
 import DropDown from './dropdown';
 
@@ -756,7 +757,7 @@ class Combobox extends React.Component {
         ];
         let hasMoreResults = false;
         let matches = new Set();
-        const searchOptions = _.uniq(this.options, 'value');
+        const searchOptions = uniqBy(this.options, 'value');
 
         for (let i = 0; i < matchRegexes.length; i++) {
             if (matches.size < this.props.maxSearchResults) {


techievivek (approved):
LGTM. I just wanted to see the values inside `this.options` and will have a look at it now. I was setting up as you suggested.

bondydaa:
can you maybe show the network tab and how long the request takes with and without your changes so we can see more firm data on how much of a performance boost this gives us?

aldo-expensify:
This shouldn't affect "how long the request takes". You can replicate the performance issue by using a code like this to generate an array and time the unique function:

```js
const getRanHex = size => {
  let result = [];
  let hexRef = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'];

  for (let n = 0; n < size; n++) {
    result.push(hexRef[Math.floor(Math.random() * 16)]);
  }
  return result.join('');
};

const generateRandomArray = (size, hexSize) => {
    const array = [];
    for (let i = 0; i < size; i++) {
        array.push({
            value: getRanHex(hexSize),
            index: i,
        });
    }
    return array;
};

const timeFunc = callback => {
    const startTime = performance.now();
    callback();
    const endTime = performance.now();
    console.log(`That took: ${endTime - startTime} milliseconds`);
};

const testArray = generateRandomArray(300000, 10);

```

then, having lodash as `_` in the scope you can:

```js
timeFunc(() => _.uniqBy(testArray, 'value'));
> That took: 196.09999990463257 milliseconds
```
if you do the same exercise, but now with underscore as `_` in the score (in www.expensify.com.dev for example), it just freezes

```js
timeFunc(() => _. uniq(testArray, 'value'));
```

bondydaa:
> This shouldn't affect "how long the request takes".

🤦 oh duh righto, still coming off vacation brain.

also nice TIL `performance` https://developer.mozilla.org/en-US/docs/Web/API/Performance/now.

but nice i think that's good enough what you posted, 🙇 thanks.

PR #445, opened 2022-05-13
Add method to force a cached code version update

alex-mechler:
### Fixed Issues
Part of https://github.com/Expensify/Expensify/issues/210381

# Tests
Test with https://github.com/Expensify/Web-Expensify/pull/33768

# QA
N/a, taken care of with https://github.com/Expensify/Web-Expensify/pull/33768

### DIFF
--- a/lib/API.jsx
+++ b/lib/API.jsx
@@ -201,6 +201,15 @@ export default function API(network, args) {
         },
 
         /**
+         * Forces an update of the stored latest code version without firing off an additional POST request
+         *
+         * @returns {exports.Deferred}
+         */
+        updateCodeVersion() {
+            return isRunningLatestVersionOfCode();
+        },
+
+        /**
          * Used to extend an API instance with new methods.
          *
          * @example


amyevans (commented):
Hmm to be honest I feel like I'm missing the point of this function - how does it avoid firing off another POST request? (Feel free to get another reviewer too with more familiarity, I don't want to hold up deploy, but I also don't feel comfortable atm signing off on it 😅)

alex-mechler:
`[hasChangedCode](https://github.com/Expensify/expensify-common/blob/9b43a7d36ba3bdaafeec6800f66fbad6d54c6434/lib/API.jsx#L199-L201)` checks a stored variable that we set in `[isRunningLatestVersionOfCode](https://github.com/Expensify/expensify-common/blob/9b43a7d36ba3bdaafeec6800f66fbad6d54c6434/lib/API.jsx#L44)`, a private method that we only ever call from `[preformPostRequest](https://github.com/Expensify/expensify-common/blob/9b43a7d36ba3bdaafeec6800f66fbad6d54c6434/lib/API.jsx#L102)` when passing an optional param. This is the only time the method was called. This PR just adds a public way to run `isRunningLatestVersionOfCode`, so that we can fire off only a GET to `revision.txt` instead of a GET to `revision.txt` and a POST to an endpoint

amyevans (approved):
🎉

PR #447, opened 2022-05-18
[No QA] Add polyfill for `replaceAll()` to fix Safari iOS 13

marcaaron:
### Fixed Issues (Related to)
$ https://github.com/Expensify/App/pull/8896#issuecomment-1130401493

# Tests

Can be tested in connection with -> https://github.com/Expensify/App/pull/9060

# QA
No QA

### DIFF
--- a/lib/str.js
+++ b/lib/str.js
@@ -2,6 +2,7 @@
 import _ from 'underscore';
 import {AllHtmlEntities} from 'html-entities';
 import {CONST} from './CONST';
+import replaceAll from 'string.prototype.replaceall';
 
 const Str = {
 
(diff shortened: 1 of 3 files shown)


lib/str.js
```diff
@@ -1049,6 +1050,18 @@ const Str = {
...
+
+    /**
+     * Polyfill for String.prototype.replaceAll
+     *
+     * @param {String} text
+     * @param {String|RegExp} searchValue
+     * @param {String|Function} replaceValue
+     * @returns {String}
+     */
+    replaceAll(text, searchValue, replaceValue) {
```
techievivek, line 1062: Just kind of curious why we have to import a package for this, why can't we use inbuilt [replaceAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll)?

marcaaron: Mostly just to backfill support for browsers where it's not available

https://caniuse.com/?search=replaceAll

marcaaron: We could implement a polyfill in every repository, but I think it is just easier in this case to add it to our "string utility library" which several other repos are using.

techievivek: Thanks for clearing this out I can see ios Safari `3.2 - 13.3` is not supported. 🙌

techievivek (approved):
Looks good.

PR #468, opened 2022-10-10
Add h1 markdown

cristipaval:
Part of [this DesignDoc](https://docs.google.com/document/d/1YJmMi9NjnstCAP2Ibo_IZdT2T16Res2jdhWr2WSrk50).
Adds support for heading markdown. A line which starts with # and a space (`# `) is wrapped with `<h1></h1>` tag.

### Fixed Issues
https://github.com/Expensify/Expensify/issues/225615

# Tests
- there are 2 new automated tests added which checks that markdown is replaced with html tags and vice versa.

- Link the `expensify-common` local repository by following [these steps](https://docs.google.com/document/d/1YJmMi9NjnstCAP2Ibo_IZdT2T16Res2jdhWr2WSrk50)
- Run the app
- Open any chat and send a text message containing lines starting with a # and a space and then text.
- Check that the lines from the previous step are rendered as headings
- copy the message to the clipboard and paste it into the composer (input text area)
- check that the message is parsed correctly and that the headings are pasted as lines starting with # and space

# QA
- Run the app
- Open any chat and send a text message containing lines starting with a # and a space and then text.
- Check that the lines from the previous step are rendered as headings
- copy the message to the clipboard and paste it into the composer (input text area)
- check that the message is parsed correctly and that the headings are pasted as lines starting with # and space

- Sending messages with all the markdowns should be tested to check that the markdown rules are applied correctly in different markdown combinations
- Copy-Paste the above messages to check that they are correctly converted back from html to markdown

### DIFF
--- a/__tests__/ExpensiMark-Markdown-test.js
+++ b/__tests__/ExpensiMark-Markdown-test.js
@@ -32,7 +32,7 @@ test('Test strikethrough HTML replacement', () => {
 });
 
 test('Test Mixed HTML strings', () => {
-    const rawHTMLTestStartString = '<em>This is</em> a <strong>test</strong>. None of <h1>these strings</h1> should display <del>as</del> <div>HTML</div>.';
+    const rawHTMLTestStartString = '<em>This is</em> a <strong>test</strong>. None of <h2>these strings</h2> should display <del>as</del> <div>HTML</div>.';
     const rawHTMLTestReplacedString = '_This is_ a *test*. None of \nthese strings\n should display ~as~ \nHTML\n.';
     expect(parser.htmlToMarkdown(rawHTMLTestStartString)).toBe(rawHTMLTestReplacedString);
 });
@@ -412,7 +412,7 @@ test('map div with bold and italics', () => {
 });
 
 test('map div with mixed html strings', () => {
-    const testString = '<div><em>This is</em> a <strong>test</strong>. None of <h1>these strings</h1> should display <del>as</del><div>HTML</div><div></div><em>line 3</em></div>';
+    const testString = '<div><em>This is</em> a <strong>test</strong>. None of <h2>these strings</h2> should display <del>as</del><div>HTML</div><div></div><em>line 3</em></div>';
     const resultString = '_This is_ a *test*. None of \nthese strings\n should display ~as~\nHTML\n_line 3_';
     expect(parser.htmlToMarkdown(testString)).toBe(resultString);
 });
--- a/__tests__/ExpensiMark-Markdown-test.js
+++ b/__tests__/ExpensiMark-Markdown-test.js
@@ -466,3 +466,17 @@ test('Test heading1 markdown replacement', () => {
     + 'this line has no special therefore it cannot be a heading1';
     expect(parser.replace(testString)).toBe(resultString);
 });
+
+test('Test html string to heading1 markdown', () => {
+    const testString = '<h1>This is a heading1 because starts with # followed by a space</h1><br />'
+    + '<h1>This is also a heading1 because starts with # followed by a space</h1><br />'
+    + 'this is not a heading1 because does not start with #<br />'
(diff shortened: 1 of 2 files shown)


cristipaval:

lib/ExpensiMark.js
```diff
@@ -162,6 +162,11 @@ export default class ExpensiMark {
                 },
                 replacement: g1 => `<blockquote>${g1}</blockquote>`,
             },
+            {
+                name: 'heading1',
+                regex: /^#(?!#) +(.*)$/gm,
```
arosiclair, line 167:
```suggestion
                regex: /^#(?!#) +(.+)$/gm,
```

I think this should only match when we have at least one character after the `# `. Otherwise this might get translated into just `<h1></h1>` which is an invalid comment in the backend. [See what I found here for more](https://github.com/Expensify/App/issues/8720#issuecomment-1105771905)

I think we should also add a test for this to ensure parsing "# " just outputs to "# ".

__tests__/ExpensiMark-Markdown-test.js
```diff
@@ -452,3 +452,35 @@ test('map real message with quotes', () => {
     const resultString = '\n> hi\n\n\n> hi\n\n';
     expect(parser.htmlToMarkdown(testString)).toBe(resultString);
 });
+
+test('Test heading1 markdown replacement', () => {
```
arosiclair, line 456: I think you should split these tests up. For example this one is testing 

- normal `# Heading` case 
- cases with multiple spaces (`#   Heading`) 
- cases with none (`#Heading`). 

You can split this into 3 small and simple tests that'll be easier to read. Also this way there's no need for the long message in the test string that explains everything 😅

arosiclair (changes requested):

lib/ExpensiMark.js
cristipaval: yeah good catch! Thanks Andrew!

__tests__/ExpensiMark-Markdown-test.js
cristipaval: yeah I like your suggestion 👍

Gonals:
Heyhey! Heads up that this PR caused [this issue](https://github.com/Expensify/App/issues/15372). We basically forgot to handle a case where the # was followed by two spaces and a new line

lib/ExpensiMark.js
```diff
@@ -223,6 +228,11 @@ export default class ExpensiMark {
                 regex: /<br(?:"[^"]*"|'[^']*'|[^'"><])*>\n?/gi,
                 replacement: '\n',
             },
+            {
+                name: 'heading1',
+                regex: /\s*<(h1)(?:"[^"]*"|'[^']*'|[^'">])*>(.*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))\s*/gi,
```
eVoloshchak, line 233: 👋
This has caused a regression in https://github.com/Expensify/App/issues/17998. 
We were removing all of the new line breaks before `<h1>` tag, (\s), but should remove whitespaces ([^\S\r\n]).
More details [here](https://github.com/Expensify/App/issues/17998#issuecomment-1524741367)

lib/ExpensiMark.js
```diff
@@ -162,6 +162,11 @@ export default class ExpensiMark {
                 },
                 replacement: g1 => `<blockquote>${g1}</blockquote>`,
             },
+            {
+                name: 'heading1',
+                regex: /^#(?!#) +(.+)$/gm,
+                replacement: '<h1>$1</h1>',
+            },
```
s77rt, line 169: This rule was misplaced. It should have been placed before quote rule, because the quote rule consumes the newlines chars i.e. `\n`. This made such markdown fail `> a\n# b`. (Coming from https://github.com/Expensify/App/issues/21846)

PR #474, opened 2022-11-22
Fix new lines when replacing <blockquote> and <br />

aldo-expensify:
This is part of a group of changes to preserving new lines when converting from `message[0].html` using `ExpensiMark.htmlToText`. This should improve:
- Copy message content to clipboard (Copy text using the clipboard button, and past in text editor: the line breaks are missing. Related issue: https://github.com/Expensify/App/issues/10262)
- It is going to be useful for showing only the first line of a message as in the LHN (currently `<blockquote>` cause buggy spacing)

### Fixed Issues
$ https://github.com/Expensify/App/issues/12894 (partially)
https://github.com/Expensify/App/issues/11494 (partially)

# Tests

Make sure unit test run (`npm run test`)

1. Go to a chat
2. Go offline
3. Compose a message by:
    1. Add an emoji
    2. Add two or more empty lines
    4. Add an emoji
    image
4. Send the message
5. Find report action in Onyx and make sure that the line breaks appear as `\n` in `message[0].text`
    image
7. Compose another message with the content:
```
before quotes
> quoted text
after quotes
```
5. Find report action in Onyx and make sure that the `message[0].text` contains `before quotes\nquoted text\nafter quotes`. This is because `<blockquote>` are block elements and they should cause line breaks in the content.
    image

# QA

1. Go to a chat
2. Go offline
3. Compose a message by:
    1. Add an emoji
    2. Add two or more empty lines
    4. Add an emoji
    image
4. Send the message
5. Find report action in Onyx and make sure that the line breaks appear as `\n` in `message[0].text`
    image
7. Compose another message with the content:
```
before quotes
> quoted text
after quotes
```
5. Find report action in Onyx and make sure that the `message[0].text` contains `before quotes\nquoted text\nafter quotes`. This is because `<blockquote>` are block elements and they should cause line breaks in the content.
    image


aldo-expensify:
cc @mollfpr since you added this code that transform html into text using regexes. Just tagging you in case you notice that what I'm doing here will cause a bug.

lib/ExpensiMark.js
```diff
@@ -287,7 +287,7 @@ export default class ExpensiMark {
         this.htmlToTextRules = [
             {
                 name: 'breakline',
-                regex: /((<br[^>]*>)+)/gi,
+                regex: /<br[^>]*>/gi,
```
mollfpr, line 290: I just roll back to the issue and the PR. Back then the issue is about the last message text in the LHN when sending a blockquote. So the solution I propose is to replace the blockquote closing tag with a space. Other regex rules in the `htmlToText` rule from the E/App that I move to expensi-mark.

I believe this will affect the text in LHN where it should be replacing adjoin break lines with a space.

luacmartins (approved):
LGTM just waiting on @mollfpr confirmation on any new issues with blockquotes

lib/ExpensiMark.js
aldo-expensify: > I believe this will affect the text in LHN where it should be replacing adjoin break lines with a space.

I had this thought after I stopped working hehe, but I just checked and it doesn't affect in web:

image

I'm going to check Android/IOS

aldo-expensify: Also, I noticed that the server:
- for the report action, it replaces each `<br>` for a `' '` (whitespace), but
     image
- for the report, the `lastMessageText` has collapsed consecutive `<br>` into a single `' '`
     image

mollfpr:
Okay, I couldn't find any issue in the current LHN with this change but some things broke the current LHN.

As you can see the `alternateText` creates spaces between the text because we replaced every breakline with a space. The `alternateText` value itself is from the `lastMessageText` key from the report object where the text is using `htmlToText`. But apparently, the current LHN doesn't show that `lastMessageText` is in LHN.
Screen Shot 2022-11-23 at 01 06 22

---

I found another issue in LHN, but it's not related to these changes but related to the `htmlToText` function. The optimistic last message in LHN should be `Confusion This`.

https://user-images.githubusercontent.com/25520267/203390527-b98abdef-01d7-4a18-b72f-a988d5d3482d.mov

---

With the current LHN behavior to show the last message, this PR will not affect the LHN but I think we should fix the above issue to show the last message correctly.

lib/ExpensiMark.js
mollfpr: I guess we should separate the function to convert the message HTML to text between LHN and chat.

aldo-expensify: Yeah, something like that... I was looking at the backend, and convert the `html` to `lastMessageText` using a function that applies the same regexes that `htmlToText` apply, including `/((<br[^>]*>)+)/`.

I was thinking that it doesn't make much sense that `🥶 <br /><br /><br />🥶` gets converted to `🥶  🥶` just collapsing the `<br />` into a single space... why the result is not `🥶 🥶` (single space) if our intention is really to collapse multiple white spaces. I'm thinking that `htmlToText` should maybe be changed to:
- Step 1: replace each `<br />` by single spaces
- Step 2: collapse all consecutive spaces into a single space

I still have to look at what the back end is doing to produce the `text` of the message in the report action from the `html`. Is there a reason why we don't collapse the spaces when deriving `text` from `html`?

aldo-expensify: The `html` of a report action's message gets converted to `text` [here](https://github.com/Expensify/Web-Expensify/blob/7e7c1eb933d5688fbd457211ca104b89feaa6f9b/lib/ReportUtils.php#L1691), and it replaces each `<br />` with a single whitespace (no collapsing).

The `lastMessageText` of the report gets derived from `html` [here](https://github.com/Expensify/Web-Expensify/blob/7e7c1eb933d5688fbd457211ca104b89feaa6f9b/lib/ReportUtils.php#L2285-L2288), collapsing multiple `<br />` into a single whitespace.

aldo-expensify: > I guess we should separate the function to convert the message HTML to text between LHN and chat.

@mollfpr so.. I think the best way to go here is:

- `htmlToText` won't collapse multiple `<br />` into a single whitespace because that is only needed for the `lastMessageText` which is used in the LHN (what this PR is doing)
- The function [ReportUtils.formatReportLastMessageText](https://github.com/Expensify/App/blob/00d7c619cc409b96e6dddca5cc0ad406f0785331/src/libs/ReportUtils.js#L396-L398) will have the responsibility of collapsing consecutive whitespaces into a single one. We already use this function to process what we put in `lastMessageText`

mollfpr: @aldo-expensify  sounds good to me!

I found an issue when testing this PR, but it's not related to the PR changes. The last message on LHN flinched without whitespace when sending a text with a blockquote.

https://user-images.githubusercontent.com/25520267/204142869-a6763984-fcb3-4433-90e5-0583b0532ddd.mov

mollfpr: Logged issue here https://expensify.slack.com/archives/C049HHMV9SM/p1669564594408699

aldo-expensify:
@mollfpr @luacmartins can you review from 0 again guys? I changed this PR to:

- `<blockquote>` are replaced by `\n` if there is content before
- `</blockquote>` are replaced by `\n` if there is content after
- `<br />` are replaced by `\n`

mollfpr:
Will test in a couple of hours.

aldo-expensify:
> Sorry, I have a question I'm a little bit confused. Is this PR will fix [Expensify/App#12894](https://github.com/Expensify/App/issues/12894) and [Expensify/App#11494](https://github.com/Expensify/App/issues/11494)?
> 

Partially, there is a PR for the API and also for the App

mollfpr (approved):

luacmartins (approved):
LGTM!

PR #484, opened 2022-12-21
Don't require password as a param for SignIn

NikkiWines:
cc: @johnmlee101 

With our new authentication logic we don't always send passwords across when signing in, so remove that param as a requirement. 
 
### Fixed Issues
$ Tied to https://github.com/Expensify/Expensify/issues/249359

# Tests
Tested in https://github.com/Expensify/Web-Expensify/pull/35889

# QA
Tested in https://github.com/Expensify/Web-Expensify/pull/35889

### DIFF
--- a/lib/API.jsx
+++ b/lib/API.jsx
@@ -354,13 +354,14 @@ export default function API(network, args) {
         /**
          * @param {Object} parameters
          * @param {String} parameters.email
-         * @param {String} parameters.password
+         * @param {String} [parameters.password]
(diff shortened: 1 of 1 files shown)


Luke9389 (approved):

Luke9389:
Can I merge this now, or is it waiting on anything?

NikkiWines:
👍 Shouldn't be waiting on anything, since it still supports passing a password, just doesn't require it any longer.

PR #493, opened 2023-01-31
Remove Subscription Change Options

Luke9389:
### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/259131

# Tests / Web QA
1. Sign into any account
2. Go to Settings > Policies > Group
3. Make a new Collect Policy
4. Toggle Auto renew Off
5. Verify that the modal looks like this:
Screenshot 2023-01-31 at 4 21 04 PM

PR #495, opened 2023-02-02
Update SUBSCRIPTION_CHANGE_REASONS

MariaHCD:
Follow up to https://github.com/Expensify/expensify-common/pull/493 due to https://github.com/Expensify/expensify-common/pull/493#issuecomment-1413678893

cc: @Luke9389 @flaviadefaria 

### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/259131

# Tests/QA
1. Sign into any account
2. Go to Settings > Policies > Group
3. Make a new Collect Policy
4. Toggle Auto renew Off
5. Verify that the options and the copy in the modal:

[Expensify - Settings - Policies.webm](https://user-images.githubusercontent.com/12268372/216336599-e2c9f4ae-5b8a-4e78-89cb-29e34add8b88.webm)

### DIFF
--- a/lib/CONST.jsx
+++ b/lib/CONST.jsx
@@ -815,22 +815,22 @@ export const CONST = {
         TOO_LIMITED: {
             id: 'tooLimited',
             label: 'Functionality needs improvement or is confusing',
-            prompt: 'What features would you like to see added or improved?',
+            prompt: 'What software are you migrating to and what led to this decision?',
         },
         TOO_EXPENSIVE: {
             id: 'tooExpensive',
             label: 'Too expensive',
-            prompt: 'What is the main reason you think Expensify is too expensive, and what do you think is a fair price?',
+            prompt: 'What software are you migrating to and what led to this decision?',
         },
         INADEQUATE_SUPPORT: {
             id: 'inadequateSupport',
             label: 'Inadequate customer support',
-            prompt: 'What about our customer support didn\'t work well for you?',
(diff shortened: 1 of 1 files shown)


lib/CONST.jsx
```diff
@@ -814,23 +814,23 @@ export const CONST = {
...
-            label: 'Functionality needs improvement or is confusing',
-            prompt: 'What features would you like to see added or improved?',
+            label: 'Functionality needs improvement',
+            prompt: 'What software are you migrating to and what led to this decision?',
         },
         TOO_EXPENSIVE: {
             id: 'tooExpensive',
             label: 'Too expensive',
-            prompt: 'What is the main reason you think Expensify is too expensive, and what do you think is a fair price?',
+            prompt: 'What software are you migrating to and what led to this decision?',
```
marcochavezf, line 823: Ah just to be sure, in the [description](https://github.com/Expensify/Expensify/issues/259131#issue-1564436048) we have this copy:

```suggestion
            prompt: 'What software are you migrating to and what do you think is a fair price?',
```

is that correct?

Luke9389: Check out the comment that's linked in the description. I think this is correct

Luke9389 (approved):

lib/CONST.jsx
Luke9389: https://github.com/Expensify/expensify-common/pull/493#issuecomment-1413689664

PR #508, opened 2023-03-08
Fixing inconsistencies between emails and autoEmail

abdulrahuman5196:
@cristipaval @Santhosh-Sellavel

### Fixed Issues
$ https://github.com/Expensify/App/issues/15621#issuecomment-1453778652

# Tests
What unit/integration tests cover your change? What autoQA tests cover your change?
Added linked email cases to existing email test cases for below
        Test markdown replacement for emails wrapped in bold/strikethrough/italic in multi-line
        Test markdown replacement for emails wrapped in bold/strikethrough/italic in a single line
        Test markdown replacement for emails containing bold/strikethrough/italic

What tests did you perform that validates your changed worked?
<details>
<summary>Web</summary>

https://user-images.githubusercontent.com/46707890/223851322-e0905269-ac20-4d16-b2f8-6bb0094a14bc.mp4

</details>

<details>
<summary>Mobile Web - Chrome</summary>

https://user-images.githubusercontent.com/46707890/223851345-f6eda165-366c-40ac-b6a9-135d6d249896.mp4

</details>

<details>
<summary>Mobile Web - Safari</summary>

https://user-images.githubusercontent.com/46707890/223851361-06efeccc-8d90-4a25-9ba0-86f650e887ae.mp4

</details>

<details>
<summary>Desktop</summary>

https://user-images.githubusercontent.com/46707890/223851425-6c382cc3-6195-4d7e-a448-fd46305f31fe.mp4

</details>

<details>
<summary>iOS</summary>

https://user-images.githubusercontent.com/46707890/223851455-e22dabdc-7e47-46fe-a0a5-77458e417ddd.mp4

</details>

<details>
<summary>Android</summary>

https://user-images.githubusercontent.com/46707890/223851482-e6ffa5e7-dae3-4189-bbef-fef1141d68c1.mp4

</details>

# QA
Input the messages
Input:
```
[test](~abc@gmail.com)
```
Result:
Only 'abc@gmail.com' should be recognised as emailId

Input:
```
[test](*abc@gmail.com)
```
Result:
Only 'abc@gmail.com' should be recognised as emailId

Input:
```
[test](abc@gmail.com)
```
Result:
The message should be recognised as 'test' with emailId as link

Areas involving emailIds


lib/Email.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable import/prefer-default-export */
```
abdulrahuman5196, line 1: Suppressing the prefer-default-import rule as this file is expected grow

danieldoglas:
@Santhosh-Sellavel please add a comment here so you can be added as a reviewer.

Santhosh-Sellavel (commented):
Adding myself as reviewer

lib/Email.js
```diff
@@ -0,0 +1,10 @@
...
+/**
+ * Email.js should be used for any common logic specific to emails.
+ */
+
+const EMAIL_REGEX = "([a-zA-Z0-9.!#$%&'+/=?^`{|}-][a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]*@[a-zA-Z0-9-]+?(\\.[a-zA-Z]+)+)";
+
+export {
+    EMAIL_REGEX,
+};
```
Santhosh-Sellavel, line 10: Not sure why should we keep this here can we move CONST?  @cristipaval

abdulrahuman5196: We already have this pattern for url links - https://github.com/Expensify/expensify-common/blob/main/lib/Url.js
So I followed the same pattern to create a file for email validations to be done.

abdulrahuman5196: URL - https://github.com/Expensify/expensify-common/blob/main/lib/ExpensiMark.js#L76

abdulrahuman5196: @Santhosh-Sellavel

Santhosh-Sellavel: There are a few other options there and some computations are done. That's not the case here may be a const variable at the top should be good or should be moved to CONST.jsx.

cc: @cristipaval What do you think?

abdulrahuman5196: Yes. I agree. I also created this file in the notion of more email features alone getting added to this file and that's why added this comment as well on "Email.js should be used for any common logic specific to emails.".
But if we strongly think new file is not required and move the same to CONST, I am ok to that as well.

abdulrahuman5196:
**This can be an additional code improvement but not mandatory requirement for this bug fix**
FYI: It seems there is an another old out-dated email regex(**blame shows 4 years**) in the common repository here - https://github.com/Expensify/expensify-common/blob/main/lib/CONST.jsx#L3.
It doesn't seem to have any usage other than right click action of email message which should be fixed by this issue - https://github.com/Expensify/App/issues/15620
Can we mark that regex as deprecated? And to only use our new common email regex for future.

Santhosh-Sellavel:
Not required I guess, because the change we are doing is to be part of markdown i.e for messages formatting in New dot. This would not be used for Email Validation.

abdulrahuman5196:
I agree. Thank you for confirmation @Santhosh-Sellavel. 

Then only action item is "Do we move the new regex to CONST or follow markdown URLs pattern"? 
Now, I feel its better to follow the markdown URLs pattern as in this PR instead of moving the same to CONST to avoid regex confusions. But let me know if we are aligned to move the regex to different place, I can make the required changes.

lib/Email.js
cristipaval: Hey there! Good catch @Santhosh-Sellavel !
@abdulrahuman5196 - Yes, I totally see why you added in a separate file trying to make it easy scalable. On the other hand, I think we should keep consistent with the existing approach (add it to `CONST.jsx`) and create a new file in the future if it will be the case to add more functionality.

abdulrahuman5196: @cristipaval  @Santhosh-Sellavel 

Thank you for the feedback. I have moved the regex to CONST file. and update the PR as well. Kindly review and let me know if anything else is required from my end.

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -61,7 +61,7 @@ test('Test multi-line strikethrough markdown replacement', () => {
 });
 
 // Emails containing *_~ are successfully wrapped in a mailto anchor tag
-test('Test markdown replacement for emails containing bold/strikethrough/italic', () => {
+test('Test markdown replacement for emails and linked emails containing bold/strikethrough/italic', () => {
```
Santhosh-Sellavel, line 64:
```suggestion
test('Test markdown replacement for emails and email links containing bold/strikethrough/italic', () => {
```

Santhosh-Sellavel (changes requested):
@cristipaval LGTM, just one comment. Can you run PR Automation checks, if all good let's merge and continue testing on PR Against E/App

abdulrahuman5196:
Thank you @Santhosh-Sellavel. Updated the minor comment.

@cristipaval  Could you kindly start the Automation Workflows? It seems I can't start it myself.

cristipaval (approved):

jjcoffee:
@abdulrahuman5196 Could you confirm why you added a new regex here for markdown emails specifically, i.e. does it do much that `EMAIL_BASE_REGEX` does not? You mention [here](https://github.com/Expensify/expensify-common/pull/508#issuecomment-1462492724) that `EMAIL_BASE_REGEX` is old, but it is used by `Str.isValidEmail` (which is used to validate emails when searching on starting a New Chat), which seems to be in sync with the backend. Would be nice if all these behaviours were consistent!

PR #521, opened 2023-04-20
avoid creating empty blockquote tag

eh2077:
### Fixed Issues
https://github.com/Expensify/App/issues/17504

# Tests
1. Go to a chat and send blank quotes, like '> \n>'
2. Verify that '> \n>' is displayed as the original text and no error shown on the comment 

https://user-images.githubusercontent.com/117511920/233407771-e1abb5b3-1174-47d0-b765-ab7a1179fe46.mov

3. Add following comment
```
> 
>quote1 line a
> quote1 line b
> 
test
> 
test
>quote2 line a
> 
> 
>quote2 line b with an empty line above
```
4. Verify blank quote is not translated and all blank rows inside a quote are retained 

https://user-images.githubusercontent.com/117511920/234045882-0577bb1e-7588-45a3-ac1e-4bf7c07359ee.mov

# QA
Same as test

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -651,14 +651,14 @@ test('Test quotes markdown replacement with quotes includes a middle blank quote
 });
 
 test('Test quotes markdown replacement with quotes includes multiple middle blank quote rows', () => {
-    const testString = '>test\n> \n> \n>test';
-    const resultString = '<blockquote>test<br /><br />test</blockquote>';
+    const testString = '>test\n> \n> \n>test\ntest\n>test\n> \n> \n> \n>test';
+    const resultString = '<blockquote>test<br /><br /><br />test</blockquote>test<br /><blockquote>test<br /><br /><br /><br />test</blockquote>';
     expect(parser.replace(testString)).toBe(resultString);
 });
 
 test('Test quotes markdown replacement with text includes blank quotes', () => {
     const testString = '> \n>quote1 line a\n> quote1 line b\ntest\n> \ntest\n>quote2 line a\n> \n> \n>quote2 line b with an empty line above';
-    const resultString = '<blockquote>quote1 line a<br />quote1 line b</blockquote>test<br />&gt; <br />test<br /><blockquote>quote2 line a<br /><br />quote2 line b with an empty line above</blockquote>';
+    const resultString = '<blockquote>quote1 line a<br />quote1 line b</blockquote>test<br />&gt; <br />test<br /><blockquote>quote2 line a<br /><br /><br />quote2 line b with an empty line above</blockquote>';
     expect(parser.replace(testString)).toBe(resultString);
 });
 
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -620,7 +620,7 @@ test('Test quotes markdown replacement and removing <br/> from <br/><pre> and </
     expect(parser.replace(testString)).toBe(resultString);
 });
 
-test('Test quotes markdown replacement skipping blank quotes ', () => {
+test('Test quotes markdown replacement skipping blank quotes', () => {
     const testString = '> \n>';
     const resultString = '&gt; <br />&gt;';
     expect(parser.replace(testString)).toBe(resultString);
@@ -629,6 +629,30 @@ test('Test quotes markdown replacement skipping blank quotes ', () => {
 test('Test quotes markdown replacement with text starts with blank quote', () => {
     const testString = '> \ntest';
     const resultString = '&gt; <br />test';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test quotes markdown replacement with quotes starts with blank quote row', () => {
+    const testString = '> \n>test';
+    const resultString = '<blockquote>test</blockquote>';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test quotes markdown replacement with quotes ends with blank quote rows', () => {
+    const testString = '>test\n> \n>';
+    const resultString = '<blockquote>test</blockquote>';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test quotes markdown replacement with quotes includes a middle blank quote row', () => {
+    const testString = '>test\n> \n>test';
+    const resultString = '<blockquote>test<br /><br />test</blockquote>';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test quotes markdown replacement with quotes includes multiple middle blank quote rows', () => {
+    const testString = '>test\n> \n> \n>test';
+    const resultString = '<blockquote>test<br /><br />test</blockquote>';
     expect(parser.replace(testString)).toBe(resultString);
 });
 
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -620,6 +620,24 @@ test('Test quotes markdown replacement and removing <br/> from <br/><pre> and </
     expect(parser.replace(testString)).toBe(resultString);
 });
 
+test('Test quotes markdown replacement skipping blank quotes ', () => {
+    const testString = '> \n>';
+    const resultString = '&gt; <br />&gt;';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test quotes markdown replacement with text starts with blank quote', () => {
+    const testString = '> \ntest';
+    const resultString = '&gt; <br />test';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test quotes markdown replacement with text includes blank quotes', () => {
+    const testString = '> \n>quote1 line a\n> quote1 line b\ntest\n> \ntest\n>quote2 line a\n> \n> \n>quote2 line b with an empty line above';
+    const resultString = '<blockquote>quote1 line a<br />quote1 line b</blockquote>test<br />&gt; <br />test<br /><blockquote>quote2 line a<br /><br />quote2 line b with an empty line above</blockquote>';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
 test('Single char matching', () => {
     const testString = ' *1* char _1_ char ~1~ char';
     const resultString = ' <strong>1</strong> char <em>1</em> char <del>1</del> char';
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -581,7 +581,7 @@ export default class ExpensiMark {
                 } else {
                     // Make sure we will only modify if we have Text needed to be formatted for quote
                     if (textToFormat !== '') {
-                        replacedText += this.formatWhitespaceForQuote(regex, textToFormat, replacement);
+                        replacedText += this.formatTextForQuote(regex, textToFormat, replacement);
                         textToFormat = '';
                     }
 
@@ -601,7 +601,7 @@ export default class ExpensiMark {
 
             // When loop ends we need the last quote to be formatted if we have quotes at last rows
             if (textToFormat !== '') {
-                replacedText += this.formatWhitespaceForQuote(regex, textToFormat, replacement);
+                replacedText += this.formatTextForQuote(regex, textToFormat, replacement);
             }
 
             // Replace all the blank lines between quotes, if there are only blank lines between them
@@ -614,7 +614,7 @@ export default class ExpensiMark {
     }
 
     /**
-     * Handle whitespace for the content of blockquote if the text matches the regex or else just return the original text
+     * Format the content of blockquote if the text matches the regex or else just return the original text
      *
      * @param {RegExp} regex
      * @param {String} textToCheck
@@ -622,7 +622,7 @@ export default class ExpensiMark {
      *
      * @returns {String}
      */
-    formatWhitespaceForQuote(regex, textToCheck, replacement) {
+    formatTextForQuote(regex, textToCheck, replacement) {
         if (textToCheck.match(regex)) {
             // Remove '&gt;' and trim each line of quote content
             let textToFormat = textToCheck.trim().split('\n').map(row => row.substr(4).trim()).join('\n');
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -581,7 +581,7 @@ export default class ExpensiMark {
                 } else {
                     // Make sure we will only modify if we have Text needed to be formatted for quote
                     if (textToFormat !== '') {
-                        replacedText += this.modifyTextForQuoteContent(regex, textToFormat, replacement);
+                        replacedText += this.formatWhitespaceForQuote(regex, textToFormat, replacement);
                         textToFormat = '';
                     }
 
@@ -601,7 +601,7 @@ export default class ExpensiMark {
 
             // When loop ends we need the last quote to be formatted if we have quotes at last rows
             if (textToFormat !== '') {
-                replacedText += this.modifyTextForQuoteContent(regex, textToFormat, replacement);
+                replacedText += this.formatWhitespaceForQuote(regex, textToFormat, replacement);
             }
 
             // Replace all the blank lines between quotes, if there are only blank lines between them
@@ -614,7 +614,7 @@ export default class ExpensiMark {
     }
 
     /**
-     * Modify text for Quote content and translate it into blockquote html if it matches the regex.
+     * Handle whitespace for the content of blockquote if the text matches the regex or else just return the original text
      *
      * @param {RegExp} regex
      * @param {String} textToCheck
@@ -622,7 +622,7 @@ export default class ExpensiMark {
      *
      * @returns {String}
      */
-    modifyTextForQuoteContent(regex, textToCheck, replacement) {
+    formatWhitespaceForQuote(regex, textToCheck, replacement) {
         if (textToCheck.match(regex)) {
             // Remove '&gt;' and trim each line of quote content
             let textToFormat = textToCheck.trim().split('\n').map(row => row.substr(4).trim()).join('\n');
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -627,8 +627,8 @@ export default class ExpensiMark {
             // Remove '&gt;' and trim each line of quote content
             let textToFormat = textToCheck.trim().split('\n').map(row => row.substr(4).trim()).join('\n');
 
-            // Remove leading and trailing line breaks and ensure that we will have only one blank row
-            textToFormat = textToFormat.trim().replace(/\n{2,}/g, '\n\n');
+            // Remove leading and trailing line breaks
+            textToFormat = textToFormat.trim();
             return replacement(textToFormat);
         }
         return textToCheck;
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -163,7 +163,7 @@ export default class ExpensiMark {
                 // inline code blocks. A single prepending space should be stripped if it exists
                 process: (textToProcess, replacement) => {
                     const regex = new RegExp(
-                        /\n?^&gt; ?(?![^<]*(?:<\/pre>|<\/code>))([^\v\n\r]+)\n?/gm,
+                        /\n?^&gt; *(?! )(?![^<]*(?:<\/pre>|<\/code>))([^\v\n\r]+)\n?/gm,
                     );
                     return this.modifyTextForQuote(regex, textToProcess, replacement);
                 },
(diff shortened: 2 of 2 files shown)


eh2077:
> @eh2077 You should have linked the issue instead of the comment with the [Proposal](https://github.com/Expensify/App/issues/17504#issuecomment-1512404819), so that Melvin knows who to assign as reviewers. All good for now, I manually added the right reviewers, but please update the link to the **[Fixed Issue](https://github.com/Expensify/expensify-common/pull/521#issue-1676915231)** and remove the $ symbol in front of the link, since this is a partial fix for it. The fix will be complete when you link the new library version to the App.

@cristipaval Thanks for the advice! I got it and and updated the PR description.

lib/ExpensiMark.js
```diff
@@ -628,4 +612,25 @@ export default class ExpensiMark {
         }
         return replacedText;
     }
+
+    /**
+     * Modify text for Quote content and translate it into blockquote html if it matches the regex.
```
yuwenmemon, line 617: Can we make this doc and method name more descriptive? Maybe call it `formatWhitespaceForQuote`?

yuwenmemon (changes requested):
Looking good and thanks a bunch for those tests! Just one small change on my end

lib/ExpensiMark.js

eh2077:
Hi @s77rt, can help to review this PR when you have time?

lib/ExpensiMark.js
```diff
@@ -628,4 +612,25 @@ export default class ExpensiMark {
...
+    /**
+     * Handle whitespace for the content of blockquote if the text matches the regex or else just return the original text
+     *
+     * @param {RegExp} regex
+     * @param {String} textToCheck
+     * @param {Function} replacement
+     *
+     * @returns {String}
+     */
+    formatWhitespaceForQuote(regex, textToCheck, replacement) {
```
s77rt, line 625: Can you explain why this change is required? What would be wrong if we just change the regex and keep everything as is?

eh2077: The regex is only used to check if the input text contains quotes, see [here](https://github.com/Expensify/expensify-common/blob/be625f55058111793bbcfe3f64788ea0a8f9d705/lib/ExpensiMark.js#L555-L561). It'll skip to translate input '> \n>', but will still create empty `blockquote` tags for input text like
```
> a
test
> 
> 
```
See below videos for comparison
**Before** 

https://user-images.githubusercontent.com/117511920/233652801-26207fb7-ce1c-4611-9739-e2b82c09e86f.mov

**After** 

https://user-images.githubusercontent.com/117511920/233652982-69214585-ee91-4ba0-b739-91c8bc39023b.mov


I think we need to avoid creating all empty `blockquote` tags, so I made this change to ensure consistency. Do you agree with this?

s77rt: Right, that would be great. But can't we do it by just altering the logic here https://github.com/Expensify/expensify-common/blob/be625f55058111793bbcfe3f64788ea0a8f9d705/lib/ExpensiMark.js#L574-L589. And without the use of an additional regex :sweat_smile:.

eh2077: @s77rt Thanks for the comment.

It should be feasible to achieve this by modifying method `modifyTextForQuote` technically but I feel it's tricky to handle 
1. blank line within a multiple line quote. Note that we only keep one blank line inside a quote between two rows.
2. trim spaces for each row of a quote
3. retain original text for empty quote, like '> \n> '

We need to find a way to retain original text, like for case '> \n> ', but the current implementation pollute `textSplitted[i]` directly and depends on the value of `textSplitted[I]` to handle blank row inside a quote.
So I think it'll be better to let `textToFormat` save the original text and extract the logic to format a quote into another method. Then both method  `modifyTextForQuote` and `formatWhitespaceForQuote` are more readable.

As for the regex, I think it's also necessary to improve it to make it consistent with the implementation of method `formatWhitespaceForQuote`

s77rt: @eh2077 Thanks for the explanation. I will have to check again to see if we can improve that. Maybe we can relay on the next line `textSplitted[i + 1]` to decide whether the current `>` should be kept as is or considered as a blockquote.

s77rt: @eh2077 I think I managed to get this working without regex. The key is to defer parsing the "tricky" lines till we manage the easy ones so it become clear if those "tricky" lines should be parsed as part of the quote or not. Do you want to work on top of that? Also can you please add more tests?

<details>
<summary>diff</summary>

```diff
diff --git a/lib/ExpensiMark.js b/lib/ExpensiMark.js
index bac4b81..21a4c2a 100644
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -155,7 +155,7 @@ export default class ExpensiMark {
                 // inline code blocks. A single prepending space should be stripped if it exists
                 process: (textToProcess, replacement) => {
                     const regex = new RegExp(
-                        /\n?^&gt; ?(?![^<]*(?:<\/pre>|<\/code>))([^\v\n\r]+)\n?/gm,
+                        /\n?^&gt; *(?! )(?![^<]*(?:<\/pre>|<\/code>))([^\v\n\r]+)\n?/gm,
                     );
                     return this.modifyTextForQuote(regex, textToProcess, replacement);
                 },
@@ -550,6 +550,7 @@ export default class ExpensiMark {
     modifyTextForQuote(regex, textToCheck, replacement) {
         let replacedText = '';
         let textToFormat = '';
+        let maybeLinesContainsTextToFormat = [];
         const match = textToCheck.match(regex);
 
         // If there's matches we need to modify the quotes
@@ -565,22 +566,17 @@ export default class ExpensiMark {
                     insideCodefence = Str.contains(textSplitted[i], '<pre>');
                 }
 
-                // We only want to modify lines starting with &gt; that is not codefence
                 if (Str.startsWith(textSplitted[i], '&gt;') && !insideCodefence) {
-                    // Avoid blank lines starting with &gt;
-                    if (textSplitted[i].trim() !== '&gt;') {
-                        textSplitted[i] = textSplitted[i].substr(4).trim();
-                        textSplitted[i] = textSplitted[i].trim();
-                        textToFormat += `${textSplitted[i]}\n`;
-
-                        // We need ensure that index i-1 >= 0
-                    } else if (i > 0) {
-                        // Certificate that we will have only one blank row
-                        if (textSplitted[i - 1].trim() !== '') {
-                            textSplitted[i] = textSplitted[i].substr(4).trim();
-                            textSplitted[i] = textSplitted[i].trim();
-                            textToFormat += `${textSplitted[i]}\n`;
+                    if (textToFormat !== '' || textSplitted[i].trim() !== '&gt;') {
+                        if (maybeLinesContainsTextToFormat.length) {
+                            maybeLinesContainsTextToFormat.forEach(j => {
+                                textToFormat += `${textSplitted[j].substr(4).trim()}\n`;
+                            });
+                            maybeLinesContainsTextToFormat = [];
                         }
+                        textToFormat += `${textSplitted[i].substr(4).trim()}\n`;
+                    } else {
+                        maybeLinesContainsTextToFormat.push(i);
                     }
                 } else {
                     // Make sure we will only modify if we have Text needed to be formatted for quote
@@ -588,6 +584,11 @@ export default class ExpensiMark {
                         textToFormat = textToFormat.replace(/\n$/, '');
                         replacedText += replacement(textToFormat);
                         textToFormat = '';
+                    } else if (maybeLinesContainsTextToFormat.length) {
+                        maybeLinesContainsTextToFormat.forEach(j => {
+                            replacedText += `${textSplitted[j]}\n`;
+                        });
+                        maybeLinesContainsTextToFormat = [];
                     }
 
                     // We dont want a \n after the textSplitted if it is the last row
@@ -608,6 +609,16 @@ export default class ExpensiMark {
             if (textToFormat !== '') {
                 textToFormat = textToFormat.replace(/\n$/, '');
                 replacedText += replacement(textToFormat);
+                textToFormat = '';
+            } else if (maybeLinesContainsTextToFormat.length) {
+                maybeLinesContainsTextToFormat.forEach(j => {
+                    if (j === textSplitted.length - 1) {
+                        replacedText += `${textSplitted[j]}`;
+                    } else {
+                        replacedText += `${textSplitted[j]}\n`;
+                    }
+                });
+                maybeLinesContainsTextToFormat = [];
             }
 
             // Replace all the blank lines between quotes, if there are only blank lines between them
```

</details>

cc @yuwenmemon

eh2077: @s77rt I like the idea to defer parsing quote lines and actually what I did in the PR uses the same idea but it defers parsing all lines start with '>' instead of only empty lines. I tested the diff you attached and it almost worked. I just made one additional change to replace this line
```js
textToFormat = textToFormat.replace(/\n$/, '');
```
to 
```js
textToFormat = textToFormat.trim().replace(/\n{2,}/g, '\n\n');
```
and then we can handle line breaks correctly. Below is the updated diff 

<details>
  <summary>Updated diff</summary>
  
```diff
diff --git a/lib/ExpensiMark.js b/lib/ExpensiMark.js
index 9f69183..f5e1239 100644
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -163,7 +163,7 @@ export default class ExpensiMark {
                 // inline code blocks. A single prepending space should be stripped if it exists
                 process: (textToProcess, replacement) => {
                     const regex = new RegExp(
-                        /\n?^&gt; ?(?![^<]*(?:<\/pre>|<\/code>))([^\v\n\r]+)\n?/gm,
+                        /\n?^&gt; *(?! )(?![^<]*(?:<\/pre>|<\/code>))([^\v\n\r]+)\n?/gm,
                     );
                     return this.modifyTextForQuote(regex, textToProcess, replacement);
                 },
@@ -560,6 +560,7 @@ export default class ExpensiMark {
     modifyTextForQuote(regex, textToCheck, replacement) {
         let replacedText = '';
         let textToFormat = '';
+        let maybeLinesContainsTextToFormat = [];
         const match = textToCheck.match(regex);
 
         // If there's matches we need to modify the quotes
@@ -575,29 +576,29 @@ export default class ExpensiMark {
                     insideCodefence = Str.contains(textSplitted[i], '<pre>');
                 }
 
-                // We only want to modify lines starting with &gt; that is not codefence
                 if (Str.startsWith(textSplitted[i], '&gt;') && !insideCodefence) {
-                    // Avoid blank lines starting with &gt;
-                    if (textSplitted[i].trim() !== '&gt;') {
-                        textSplitted[i] = textSplitted[i].substr(4).trim();
-                        textSplitted[i] = textSplitted[i].trim();
-                        textToFormat += `${textSplitted[i]}\n`;
-
-                        // We need ensure that index i-1 >= 0
-                    } else if (i > 0) {
-                        // Certificate that we will have only one blank row
-                        if (textSplitted[i - 1].trim() !== '') {
-                            textSplitted[i] = textSplitted[i].substr(4).trim();
-                            textSplitted[i] = textSplitted[i].trim();
-                            textToFormat += `${textSplitted[i]}\n`;
+                    if (textToFormat !== '' || textSplitted[i].trim() !== '&gt;') {
+                        if (maybeLinesContainsTextToFormat.length) {
+                            maybeLinesContainsTextToFormat.forEach(j => {
+                                textToFormat += `${textSplitted[j].substr(4).trim()}\n`;
+                            });
+                            maybeLinesContainsTextToFormat = [];
                         }
+                        textToFormat += `${textSplitted[i].substr(4).trim()}\n`;
+                    } else {
+                        maybeLinesContainsTextToFormat.push(i);
                     }
                 } else {
                     // Make sure we will only modify if we have Text needed to be formatted for quote
                     if (textToFormat !== '') {
-                        textToFormat = textToFormat.replace(/\n$/, '');
+                        textToFormat = textToFormat.trim().replace(/\n{2,}/g, '\n\n');
                         replacedText += replacement(textToFormat);
                         textToFormat = '';
+                    } else if (maybeLinesContainsTextToFormat.length) {
+                        maybeLinesContainsTextToFormat.forEach(j => {
+                            replacedText += `${textSplitted[j]}\n`;
+                        });
+                        maybeLinesContainsTextToFormat = [];
                     }
 
                     // We dont want a \n after the textSplitted if it is the last row
@@ -616,8 +617,18 @@ export default class ExpensiMark {
 
             // When loop ends we need the last quote to be formatted if we have quotes at last rows
             if (textToFormat !== '') {
-                textToFormat = textToFormat.replace(/\n$/, '');
+                textToFormat = textToFormat.trim().replace(/\n{2,}/g, '\n\n');
                 replacedText += replacement(textToFormat);
+                textToFormat = '';
+            } else if (maybeLinesContainsTextToFormat.length) {
+                maybeLinesContainsTextToFormat.forEach(j => {
+                    if (j === textSplitted.length - 1) {
+                        replacedText += `${textSplitted[j]}`;
+                    } else {
+                        replacedText += `${textSplitted[j]}\n`;
+                    }
+                });
+                maybeLinesContainsTextToFormat = [];
             }
 
             // Replace all the blank lines between quotes, if there are only blank lines between them
```
</details>

Based on the updated diff above, I think extracting the logic to format a `blockquote` tag into another method does improve the readability of the code. Do you agree with this? Or I'd like to hear your concerns about refactoring it into a new method.

Below are the test cases I tested and will be added
<details>
  <summary>Test cases</summary>

```js
test('Test quotes markdown replacement skipping blank quotes ', () => {
    const testString = '> \n>';
    const resultString = '&gt; <br />&gt;';
});

test('Test quotes markdown replacement with text starts with blank quote', () => {
    const testString = '> \ntest';
    const resultString = '&gt; <br />test';
});

test('Test quotes markdown replacement with quotes starts with blank quote row', () => {
    const testString = '> \n>test';
    const resultString = '<blockquote>test</blockquote>';
});

test('Test quotes markdown replacement with quotes ends with blank quote rows', () => {
    const testString = '>test\n> \n>';
    const resultString = '<blockquote>test</blockquote>';
});

test('Test quotes markdown replacement with quotes includes a middle blank quote row', () => {
    const testString = '>test\n> \n>test';
    const resultString = '<blockquote>test<br /><br />test</blockquote>';
});

test('Test quotes markdown replacement with quotes includes multiple middle blank quote rows', () => {
    const testString = '>test\n> \n> \n>test';
    const resultString = '<blockquote>test<br /><br />test</blockquote>';
});

test('Test quotes markdown replacement with text includes blank quotes', () => {
    const testString = '> \n>quote1 line a\n> quote1 line b\ntest\n> \ntest\n>quote2 line a\n> \n> \n>quote2 line b with an empty line above';
    const resultString = '<blockquote>quote1 line a<br />quote1 line b</blockquote>test<br />&gt; <br />test<br /><blockquote>quote2 line a<br /><br />quote2 line b with an empty line above</blockquote>';
});
```
  
</details>

s77rt: I noticed the diff in line breaks but I don't understand why we are doing it that way. The current implementation is buggy and whatever is done with line breaks seems unintended. It appears to me that the new behaviour of line breaks makes more sense and it's something that would match your expectations. Or maybe I'm missing something here?

___


Current:
```
> a
>
> b
```
One line break

```
> a
>
>
> b
```
One line breaks

```
> a
>
>
>
> b
```
Two line breaks

___


New:
```
> a
>
> b
```
One line break

```
> a
>
>
> b
```
Two line breaks

```
> a
>
>
>
> b
```
Three line breaks

eh2077: @s77rt @yuwenmemon 
I tested following quotes and attached the screenshots FYI. Please let me know if you're fine with keeping one blank row between two rows of a quote.

```
> quote 1
> 
> after 1 blank row

> quote 2
> 
> 
> after 2 blank rows

> quote 3
> 
> 
> 
> after 3 blank rows

> quote 4
> 
> 
> 
> 
> after 4 blank rows

> quote 5
> 
> 
> 
> 
> 
> after 5 blank rows
```

| new.expensify.com  | This PR |
| ------------- | ------------- |
| Screenshot 2023-04-24 at 4 24 35 PM  | Screenshot 2023-04-24 at 4 24 02 PM |

s77rt: I think we should ask on Slack. Can you please compile 3 screenshots
1. The pair line break (Staging Expensify)
2. The one line break
3. The all line break

eh2077:
@s77rt @yuwenmemon According to the Slack [discussion](https://expensify.slack.com/archives/C01GTK53T8Q/p1682331970891359), we'll keep all blank rows and I updated the PR. Please help to review it again at your convenience, thanks!

yuwenmemon (dismissed):

lib/ExpensiMark.js
```diff
@@ -628,4 +612,25 @@ export default class ExpensiMark {
...
+        if (textToCheck.match(regex)) {
```
s77rt, line 626: I thought we agreed on using a regex-free implementation :sweat_smile: Do we really still need that extra regex?

s77rt (changes requested):

lib/ExpensiMark.js
eh2077: I think it's fine to use the same regex to test the input here. It'll be consistent and explicit to determine if the input text should be translated into a quote.

s77rt: I think we should avoid that for performance reasons. Regex is slow. Using regex for every line is an overkill.

eh2077: > Using regex for every line is an overkill.

Actually, we don't call regex matching on each row but on a bunch of rows that start with `&gt;`.

I think we don't need to concern about the regex performance issue here as the regex to match quote is moderate. What I feel about this lib is that it always prefers regex(conciseness provided by regex) to string modification.

s77rt: Oh actually we only call the regex once for each blockquote. So something like
```
Test
> hi
> there
Abc
> a
> b
> c
```

The regex call here is just 2, not 5 and not 7 :sweat_smile:

That's cool, I also like how the code is much simplified, will perform some more tests and get back to you

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -620,6 +620,48 @@ test('Test quotes markdown replacement and removing <br/> from <br/><pre> and </
     expect(parser.replace(testString)).toBe(resultString);
 });
```
s77rt, line 622: We have some failing tests, please update accordingly.

lib/ExpensiMark.js
s77rt, line 625: I think the function name and doc comment can be improved. This feels to me like `formatTextForQuote` as it's not only for whitespace.

lib/ExpensiMark.js
```diff
@@ -628,4 +612,25 @@ export default class ExpensiMark {
...
+            // Remove '&gt;' and trim each line of quote content
+            let textToFormat = textToCheck.trim().split('\n').map(row => row.substr(4).trim()).join('\n');
```
s77rt, line 628: I don't see the point of trimming `textToCheck`. Can you please explain what purpose this serves?

s77rt (changes requested):

lib/ExpensiMark.js
eh2077: Because we concat `textToCheck` through
```js
textToFormat += `${textSplitted[i]}\n`;
```
So `textToCheck` ends with `\n` and to void the empty string item return by `split('\n')`, we need to trim it before split it by `\n`

s77rt: But we already trim the output in the next line

__tests__/ExpensiMark-HTML-test.js
eh2077: I already updated tests after we decided to retain all blanks rows. `npm run test` is happy on my Mac and the GH check is also green. I'm a little bit confused about the failing tests you mentioned. Can you share more details please? Maybe I'm missing something

lib/ExpensiMark.js
eh2077: Let's say we have textToCheck = '&gt; a\n', then textToCheck.split('\n') will return ['&gt; a', ''] if we don't trim it before splitting it. Hmm, we'll map `row => row.substr(4).trim()` for empty string through I just tested and found method call `''.substr(4)` won't return error.

eh2077: I'll remove it if you 👍 it

lib/ExpensiMark.js
eh2077: updated

lib/ExpensiMark.js
s77rt: I don't see much value added keeping it or removing it, let's remove just for code clarity

eh2077: updated

__tests__/ExpensiMark-HTML-test.js
eh2077: @s77rt 
That's strange. Both [GH check](https://github.com/Expensify/expensify-common/actions/runs/4795896068/jobs/8531013931?pr=521) and my client are passed.

Hmm, did you checkout this branch to test on your client? If so, can you check the diff of `git diff HEAD~6`

s77rt: Oh so sorry. I was using an outsynced repo. The br difference is due to another change unrelated to this PR. All good now

s77rt (approved):
LGTM! :rocket: 

cc @yuwenmemon

yuwenmemon (approved):

PR #522, opened 2023-04-25
allow heading inside a quote

eh2077:
### Fixed Issues
https://github.com/Expensify/App/issues/17367

# Tests
1. Send a message, `> # heading`
2. Verify the text is displayed in as header inside the quote
3. Send following message
```
========= test case 1 ========
> [quote] case 1
> # heading
> [quote] case 1

[normal] case 1
# heading
[normal] case 1

========= test case 2 ========
> [quote] case 2
> 
> # heading
> 
> [quote] case 2

[normal] case 2

# heading

[normal] case 2
```
4. Verify normal heading and heading inside quote have the same line break handling

https://user-images.githubusercontent.com/117511920/234178742-048ea7c0-5e49-47f6-8d82-3f5e78daa8d5.mov

# QA
Same as Tests above

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -264,7 +264,7 @@ export default class ExpensiMark {
                 name: 'quote',
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
-                    const resultString = g2.replace(/\[block\]/g, '\n').trim()
+                    const resultString = g2.replace(/(\[block\])+/g, '\n').trim()
                         .split('\n').map(m => `> ${m}`)
                         .join('\n');
                     return `\n${resultString}\n`;
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -167,7 +167,10 @@ export default class ExpensiMark {
                     );
                     return this.modifyTextForQuote(regex, textToProcess, replacement);
                 },
-                replacement: g1 => `<blockquote>${g1}</blockquote>`,
+                replacement: (g1) => {
+                    const replacedText = this.replace(g1, {filterRules: ['heading1'], shouldEscapeText: false});
+                    return `<blockquote>${replacedText}</blockquote>`;
+                },
             },
             {
                 name: 'heading1',
@@ -261,7 +264,10 @@ export default class ExpensiMark {
                 name: 'quote',
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
-                    const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
+                    // const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
+                    const resultString = g2.replace(/\[block\]/g, '\n').trim()
+                        .split('\n').map(m => `> ${m}`)
+                        .join('\n');
                     return `\n${resultString}\n`;
                 },
             },
@@ -298,6 +304,16 @@ export default class ExpensiMark {
                 name: 'breakline',
                 regex: /<br[^>]*>/gi,
                 replacement: '\n',
+            },
+            {
+                name: 'blockquoteWrapHeadingOpen',
+                regex: /<blockquote><h1>/gi,
+                replacement: '<blockquote>',
+            },
+            {
+                name: 'blockquoteWrapHeadingClose',
+                regex: /<\/h1><\/blockquote>/gi,
+                replacement: '</blockquote>',
             },
             {
                 name: 'blockElementOpen',
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -265,7 +265,8 @@ export default class ExpensiMark {
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
                     const resultString = g2.replace(/(\[block\])+/g, '\n').trim()
-                        .split('\n').map(m => `> ${m}`)
+                        .split('\n')
+                        .map(m => `> ${m}`)
                         .join('\n');
                     return `\n${resultString}\n`;
                 },
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -264,7 +264,6 @@ export default class ExpensiMark {
                 name: 'quote',
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
-                    // const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
                     const resultString = g2.replace(/\[block\]/g, '\n').trim()
                         .split('\n').map(m => `> ${m}`)
                         .join('\n');
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -264,7 +264,8 @@ export default class ExpensiMark {
                 name: 'quote',
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
-                    const resultString = g2.replace(/(\[block\])+/g, '\n').trim()
+                    const resultString = g2.replace(/(\[block\])+/g, '\n')
+                        .trim()
                         .split('\n')
                         .map(m => `> ${m}`)
                         .join('\n');
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -720,4 +720,7 @@ test('Test quotes markdown replacement with heading inside', () => {
 
     testString = '> test\n> # heading\n> test';
     expect(parser.replace(testString)).toBe('<blockquote>test<br /><h1>heading</h1>test</blockquote>');
+
+    testString = '> # heading A\n> # heading B';
+    expect(parser.replace(testString)).toBe('<blockquote><h1>heading A</h1><h1>heading B</h1></blockquote>');
 });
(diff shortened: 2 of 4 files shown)


pecanoro:
@Santhosh-Sellavel I think if you leave a comment I will be able to add you as a reviewer, I think the settings for this repo are not properly set so I can add external people.

Santhosh-Sellavel (commented):
Adding myself as a reviewer

hayata-suenaga (dismissed):
The test cases make sense. I'm not well versed in markdown and string manipulation. I'd wait other people's preview.

lib/ExpensiMark.js
```diff
@@ -261,7 +264,9 @@ export default class ExpensiMark {
                 name: 'quote',
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
-                    const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
+                    const resultString = g2.replace(/\[block\]/g, '\n').trim()
+                        .split('\n').map(m => `> ${m}`)
+                        .join('\n');
```
Santhosh-Sellavel, line 269: Why are we doing this here?

Santhosh-Sellavel: @eh2077 bump!

eh2077: @Santhosh-Sellavel Thanks for looking into this.

As we translated `<h1>` tag into `[block]# $2[block]` using the separator `[block]` to handle line break between markdown block elements, like heading, code fence and quote, see
https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L233-L237

Currently, we handle line break between markdown block element in method [replaceBlockWithNewLine](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L475-L501)
https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L475-L501
While inside a quote, I think it'll be better to handle the `[block]` separators which wrap the heading here in the replacement method of html-blockquote-to-markdown-quote rule because
1. We'll only have one kind of markdown block element, heading, inside a quote
2. It's not that easy to modify method [replaceBlockWithNewLine](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L475-L501) to handle heading inside a quote

eh2077: When writing the above the comment, I found another case of having two continuous headers, like
```
===== normal heading =====
# heading A
# heading B
===== heading in quote =====
> # heading A inside quote
> # heading B inside quote
```
I fixed it and added tests. See below video

https://user-images.githubusercontent.com/117511920/234743733-d01b1a95-947c-45b3-8808-3def875fce29.mov

lib/ExpensiMark.js
```diff
@@ -261,7 +264,9 @@ export default class ExpensiMark {
                 name: 'quote',
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
-                    const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
+                    const resultString = g2.replace(/(\[block\])+/g, '\n').trim()
+                        .split('\n').map(m => `> ${m}`)
+                        .join('\n');
```
Santhosh-Sellavel, line 269: @eh2077 

Why we are doing this is still unclear.

> Currently, we handle line break between markdown block element in method [replaceBlockWithNewLine](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L475-L501)

Is there any problem with that? 

Why do we have to do the following?

> While inside a quote, I think it'll be better to handle the [block] separators which wrap the heading here in the replacement method of html-blockquote-to-markdown-quote rule because|
> 
> 1. We'll only have one kind of markdown block element, heading, inside a quote
> 2. It's not that easy to modify method [replaceBlockWithNewLine](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L475-L501) to handle heading inside a quote

I'm not following what's are the case and what's the another case you are mentioning?

> When writing the above the comment, I found another case of having two continuous headers, like
> ```
> ===== normal heading =====
> # heading A
> # heading B
> ===== heading in quote =====
> > # heading A inside quote
> > # heading B inside quote
> ```

eh2077: @Santhosh-Sellavel 

> @eh2077
> 
> Why we are doing this is still unclear.
> 
> > Currently, we handle line break between markdown block element in method [replaceBlockWithNewLine](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L475-L501)
> 
> Is there any problem with that?

Yes, the method [replaceBlockWithNewLine](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L475-L501) will have a problem to handle line break for header inside a quote. Let's use html 
```html
<blockquote><h1>heading</h1>test</blockquote>
```
as an example. The html above is translated into 
```
<blockquote>[block]# heading[block]test</blockquote>
```
by [heading1](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L233-L237) rule.
Then the current implementation of [quote](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L264) rule translates the blockquote into 
```
> [block]# heading[block]test
```
Last, the above output is translated into 
```
> # heading
test
```
by method [replaceBlockWithNewLine](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L475-L501).
The first `[block]` is eliminated occasionally because of quote character `>`, see [this line](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L493). The second `[block]` is translated into a line break. See below [code snippet](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L493-L497) from method replaceBlockWithNewLine
```js
            // Insert '\n' unless it ends with '\n' or '>' or it's the last [block].
            if (text.match(/[\n|>][>]?[\s]?$/) || index === splitText.length - 1) {
                joinedText += text;
            } else {
                joinedText += `${text}\n`;
            }
```
So we need to fix the replacement of `[block]` for header inside a quote.

> 
> Why do we have to do the following?
> 
> > While inside a quote, I think it'll be better to handle the [block] separators which wrap the heading here in the replacement method of html-blockquote-to-markdown-quote rule because|
> > 
> > 1. We'll only have one kind of markdown block element, heading, inside a quote
> > 2. It's not that easy to modify method [replaceBlockWithNewLine](https://github.com/Expensify/expensify-common/blob/3cdaa947fe77016206c15e523017cd50678f2359/lib/ExpensiMark.js#L475-L501) to handle heading inside a quote

Based on the above explanation, we know that the method replaceBlockWithNewLine relies on character `>` to handle line break, which is possibly conflicted with the syntax of markdown quote. So, my idea here is to try to keep the existing logic of replaceBlockWithNewLine not changed and handle the new use case inside the replacement function of quote. 

> 
> I'm not following what's are the case and what's the another case you are mentioning?
> 
> > When writing the above the comment, I found another case of having two continuous headers, like
> > ```
> > ===== normal heading =====
> > # heading A
> > # heading B
> > ===== heading in quote =====
> > > # heading A inside quote
> > > # heading B inside quote
> > ```

Sorry, my bad I should use a more concrete example. Let's say we have two headers in a quote, like
```
> # heading A
> # heading B
```
Then it'll be translated into html
```html
<blockquote><h1>heading A</h1><h1>heading B</h1></blockquote>
```
When we edit the comment, the headers html is translated into markdown, like
```
> [block]# heading A[block][block]# heading B[block]
```
So, we need to replace `[block][block]` with one line break instead of two using following replacement
```
const resultString = g2.replace(/(\[block\])+/g, '\n').trim()
```

Santhosh-Sellavel (approved):
LGTM, @pecanoro all yours!

Will test PR against E/APP!

pecanoro (approved):
Tested well!

pecanoro:
@Santhosh-Sellavel @eh2077 Do you know why we remove the empty line after the heading in case 2 but we leave the one above the heading? I mean, that's the behavior with and without quotes but I find it pretty odd.

lib/ExpensiMark.js
```diff
@@ -261,7 +264,9 @@ export default class ExpensiMark {
                 name: 'quote',
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
-                    const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
+                    const resultString = g2.replace(/(\[block\])+/g, '\n').trim()
+                        .split('\n').map(m => `> ${m}`)
```
pecanoro, line 268: Move the `.map` no a new line as well.

pecanoro (changes requested):

eh2077:
> @Santhosh-Sellavel @eh2077 Do you know why we remove the empty line after the heading in case 2 but we leave the one above the heading? I mean, that's the behavior with and without quotes but I find it pretty odd.

@pecanoro That's an interesting question. I guessed it's just because the product team want to have line breaks before heading and want to keep content after heading look compact(so removing line breaks after heading).

lib/ExpensiMark.js

hayata-suenaga (approved):
LGTM!

eh2077:
> @Santhosh-Sellavel @eh2077 Do you know why we remove the empty line after the heading in case 2 but we leave the one above the heading? I mean, that's the behavior with and without quotes but I find it pretty odd.

@pecanoro I just encountered this again in another issue when addressing proposal review feedback from C+ team, see https://github.com/Expensify/App/issues/17998#issuecomment-1530002233. I think it's possible to improve this weird line break handling issue in that issue.

lib/ExpensiMark.js
```diff
@@ -261,7 +264,10 @@ export default class ExpensiMark {
                 name: 'quote',
                 regex: /\n?<(blockquote|q)(?:"[^"]*"|'[^']*'|[^'">])*>([\s\S]*?)<\/\1>(?![^<]*(<\/pre>|<\/code>))/gi,
                 replacement: (match, g1, g2) => {
-                    const resultString = g2.trim().split('\n').map(m => `> ${m}`).join('\n');
+                    const resultString = g2.replace(/(\[block\])+/g, '\n').trim()
```
pecanoro, line 267: We also need the trim in a new line, sorry I missed it!

pecanoro (changes requested):

lib/ExpensiMark.js
eh2077: updated

pecanoro (approved):

PR #525, opened 2023-04-28
Return early if given falsey values in SMS methods

jasperhuangg:
### Fixed Issues
https://expensify.slack.com/archives/C049HHMV9SM/p1682358290819999

# Tests
Added some unit tests


lib/str.js
```diff
@@ -942,6 +942,9 @@ const Str = {
      * @return {String}
      */
     removeSMSDomain(text) {
+        if (!text) {
+            return text;
```
neil-marcellini, line 946: `return ''` so we always return a string

neil-marcellini (changes requested):

PR #530, opened 2023-05-08
skip rendering invalid markdown

eh2077:
### Fixed Issues
$ https://github.com/Expensify/App/issues/17785

# Tests
1. Go to a chat and add a comment, `_*test_*`
2. Verify that text `test` is displayed in italic and is not formatted in bold style.

# QA
Same as test


eh2077:
Hi @danieldoglas, can you help to add @deetergp and @eVoloshchak as the reviewers of this PR? Thanks!

danieldoglas:
I was able to add @deetergp , but I think @eVoloshchak will need to do a review directly - for some reason I can't add them.

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -724,3 +724,14 @@ test('Test quotes markdown replacement with heading inside', () => {
     testString = '> # heading A\n> # heading B';
     expect(parser.replace(testString)).toBe('<blockquote><h1>heading A</h1><h1>heading B</h1></blockquote>');
 });
+
+test('Skip rendering invalid markdown',() => {
+    let testString = '_*test_*';
+    expect(parser.replace(testString)).toBe('<em>*test</em>*');
```
deetergp, line 730: I thought we decided to not do any rendering for when the markdown was invalid? Unless I'm missing something, there doesn't seem to be any rhyme or reason to why any of these appear to render they way they do.

deetergp (changes requested):

__tests__/ExpensiMark-HTML-test.js
eh2077: @deetergp Thanks for looking into this. I got your point but I think it will make more sense to parse italic style for text `_*test_` because it's a valid markdown italic style. Note that we apply the italic rule 
https://github.com/Expensify/expensify-common/blob/d9c8b08ca67363e4442291ae283326d25f445cc5/lib/ExpensiMark.js#L174-L177
first and then apply the rule of bold style. So, what we're going to do is to reject rendering when the bold rule
https://github.com/Expensify/expensify-common/blob/d9c8b08ca67363e4442291ae283326d25f445cc5/lib/ExpensiMark.js#L182-L184
matches text `*test</em>*` which is overlapping with an `em` tag. The Github works in the same way and it render following text
```md
_**test_**
_test_
**test**
```
into
_**test_**
_test_
**test**

eh2077: @deetergp friendly bump! Can I get your feedback about my last comment?

deetergp: Sorry @eh2077, it's been a crazy last couple days. I think if you're going to render any of the tags, you should go wit the first and ignore the rest.
### From
```
_*test_*
*_test*_
~*test~*
*~test*~
```

### Into
_*test_*
*_test*_
~*test~*
*~test*~

Your test for `~*test~*` shouldn't be passing, it should be wrapped in `<del>` instead of `<strong>` shouldn't it?

eh2077: > I think if you're going to render any of the tags, you should go with the first and ignore the rest.

@deetergp That's a good point. Ideally, it should be handled in this way but, actually, we render them by a fixed order, `italic` => `bold` => `strikethrough`, see
https://github.com/Expensify/expensify-common/blob/5f4672c84e19122fc77b4bf1ac092d4919472112/lib/ExpensiMark.js#L174-L190
and method ExpensiMark.replace
https://github.com/Expensify/expensify-common/blob/5f4672c84e19122fc77b4bf1ac092d4919472112/lib/ExpensiMark.js#L381-L386

That's why we render `<strong>` tag and then reject rendering `<del>` tag for input `~*test~*`.

Based on the current architecture of ExpensiMark.js, I feel it's not easy to achieve the intuitive behaviour you suggested. 

One idea comes into my mind is, instead of replacing all `<em>`, `<strong>` and `<del>` tags one by one, we can compare and replace each occurrence of them one by one. We can compare the index of `regex.exec(string)` to choose the match with smallest index to render first. For example

```js
const test = '~*test~* abc *~def*~ efg';
console.log(/\B\*((?=\S)(([^\s*]|\s(?!\*))+?))\*\B(?![^<]*(<\/pre>|<\/code>|<\/a>))/g.exec(test));
// output
// [
//     '~*test~',
//     '*test',
//     '*test',
//     't',
//     undefined,
//     index: 0,
//     input: '~*test~* abc *~def*~ efg',
//     groups: undefined
// ]
console.log(/\B~((?=\S)((~~(?!~)|[^\s~]|\s(?!~))+?))~\B(?![^<]*(<\/pre>|<\/code>|<\/a>))/g.exec(test));
// output
// [
//     '*test~*',
//     'test~',
//     'test~',
//     '~',
//     undefined,
//     index: 1,
//     input: '~*test~* abc *~def*~ efg',
//     groups: undefined
// ]
```

We can have something like a heap data structure to handle it. Sounds a little bit complicated and I'm not sure if it's worth to implement it for this issue. I'll be happy to hear your thoughts on this!

deetergp: Yes, you're right, the order is pretty set, so the results make sense. We're probably heading toward a need to refactor ExpensiMark, but that is definitely out-of-scope for this issue. For now, let's go with what you've got in this PR and move forward from there.

@eVoloshchak I have no idea what's going on, but I can't seem to add you as a reviewer for this.

deetergp (approved):

PR #543, opened 2023-06-02
Use htmlDecode instead of unescape

therealsujitk:
Currently `_.unescape()` is used to replace entities while html is being converted to text, however this function only replaces a limited number of entities resulting in only a part of the string being unescaped in a few cases. Because of this I've replaced it with `Str.htmlDecode()`.

For example `&#32;` is not unescaped by `_.unescape()` but is added in the replace function in line 36.

https://github.com/Expensify/expensify-common/blob/925e62744f784a92c3ad8947aabfc21c93bec221/lib/ExpensiMark.js#L23-L39

### Fixed Issues
$ https://github.com/Expensify/App/issues/19789

# Tests
    - Added html entities to the unescape test in `ExpensiMark-test.js`.
    - Previously `&amp;&#32;&amp;amp;&#32;&amp;lt;&#32;&lt;` was being unescaped to `&&#32;&amp;&#32;&lt;&#32;<`, now it'll be unescaped to `& &amp; &lt; <`.
    - Made sure all unit tests passed.

# QA
    - Test the function with other [HTML entities](https://www.w3schools.com/html/html_entities.asp) to verify that they are being unescaped or simply run all unit tests.
    - Directly pushing these changes to other repositories such as [Expensify/App](https://github.com/Expensify/App) may cause some regressions.
    - While pushing these changes a few other changes have to be made. There are a few occurrences of `Str.htmlDecode()` being used on top of `parser.htmlToText()`, these will have to be changed to just `parser.htmlToText()` or it may result in a double unescape.

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -587,7 +587,8 @@ export default class ExpensiMark {
         });
 
         // Unescaping because the text is escaped in 'replace' function
-        replacedText = _.unescape(replacedText);
+        // We use 'htmlDecode' instead of 'unescape' to replace entities like '&#32;'
+        replacedText = Str.htmlDecode(replacedText);
         return replacedText;
     }
 
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -23,7 +23,7 @@ export default class ExpensiMark {
             {
                 name: 'codeFence',
 
-                // &#60; is a backtick symbol we are matching on three of them before then after a new line character
+                // &#x60; is a backtick symbol we are matching on three of them before then after a new line character
                 regex: /(&#x60;&#x60;&#x60;[\n]?)((?:\s*?(?![\n]?&#x60;&#x60;&#x60;(?!&#x60;))[\S])+\s*?)((?=\n?)&#x60;&#x60;&#x60;)/g,
 
                 // We're using a function here to perform an additional replace on the content
--- a/__tests__/ExpensiMark-test.js
+++ b/__tests__/ExpensiMark-test.js
@@ -10,7 +10,7 @@ test('Test text is escaped', () => {
 });
 
 test('Test text is unescaped', () => {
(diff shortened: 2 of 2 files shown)


therealsujitk:

alex-mechler:
@therealsujitk since the related issue currently does not have an accepted proposal, and is on hold, I'm going to hold off reviewing for now

s77rt:
@alex-mechler Assign me here, will handle this as a part of https://github.com/Expensify/App/issues/19789

alex-mechler:
Assign for review or assign as assignee?

s77rt:
I don't think we have assignee for PRs. So I think I will just be a reviewer

lib/ExpensiMark.js
```diff
@@ -587,7 +587,8 @@ export default class ExpensiMark {
         });
 
         // Unescaping because the text is escaped in 'replace' function
-        replacedText = _.unescape(replacedText);
+        // We use 'htmlDecode' instead of 'unescape' to replace entities like '&#32;'
+        replacedText = Str.htmlDecode(replacedText);
```
s77rt, line 591: should we use htmlEncode the the replace function? Or just keep this change for the htmlToText function?

s77rt:
I had to add a review manually, not sure why we can't be asked for review. I think this is a known issue as we encountered this before

lib/ExpensiMark.js
therealsujitk: We can, but I don't think it's necessary. The only reason `_.unescape()` doesn't work is because we're adding `&#32;` in the `replace()` function. Otherwise it would work.

s77rt (approved):
LGTM! :rocket: 

cc @alex-mechler

alex-mechler (approved):
These changes also LGTM, but addined @stitesExpensify as a reviewer since they have more context

stitesExpensify (approved):

PR #544, opened 2023-06-06
Support italic style for @here mention

eh2077:
### Fixed Issues
$ https://github.com/Expensify/App/issues/19857

# Tests

https://github.com/Expensify/expensify-common/assets/117511920/f2a46c8e-0100-4448-8dfc-b35b1bfe9c1d

1. Add a comment `_@here_`
2. Verify that `@here` mention is displayed in italic

# QA
Same as test

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -934,6 +934,12 @@ test('Test for here mention with invalid username', () => {
     expect(parser.replace(testString)).toBe(resultString);
 });
 
+test('Test for @here mention with italic style', () => {
+    const testString = '_@here_ @here_123 @here_abc @here123 @hereabc';
+    const resultString = '<em><mention-here>@here</mention-here></em> @here_123 @here_abc @here123 @hereabc';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
 test('Test for @here mention without space or supported styling character', () => {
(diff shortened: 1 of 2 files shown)


eh2077:
@allroundexperts Updated to include more unit tests

hayata-suenaga (dismissed):
LGTM

eh2077:
> > @allroundexperts Updated to include more unit tests
> 
> @eh2077 This looks good. Can you also add tests for cases like: `<a href="">_@here_</a>` etc? I can see that you have added one test for `<a href="">@here</a>` but one with `_` would be helpful here as well. I think it is also good to add tests that test for consistency between `_@here_` and `*@here*`. This will make sure that we never go in-consistent with the two and will surely prevent regressions in the future.

@allroundexperts I updated to add test cases for `*@here*` and `~@here~` as well. 

As I mentioned in my proposal, we'll need the fix from another issue(see [proposal](https://github.com/Expensify/App/issues/19997#issuecomment-1573016165)) to not render `@here` mention for input `[_@here_](google.com)`. Note that currently, it's translated into html
```html
<a href="https://google.com" target="_blank" rel="noreferrer noopener"><em><mention-here>@here</mention-here></em></a>
```
which is not expected. So, I just decided not to include it into the test.

MonilBhavsar (approved):
Looks good!
@hayata-suenaga all yours

hayata-suenaga (approved):

PR #545, opened 2023-06-08
upgrade html-entities version

hungvu193:
### Fixed Issues
$ https://github.com/Expensify/App/issues/19790

# Tests
1. Open any chat
2. Send an invalid html character, eg: &#50000000;
3. Verify that the app isn't crashed and there's no console error.

# QA
Same as Tests.

# Result

https://user-images.githubusercontent.com/16502320/241865033-1bf8214e-8d72-4ca0-bce2-4d73a5998e88.mov

### DIFF
--- a/lib/str.js
+++ b/lib/str.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-control-regex */
 import _ from 'underscore';
-import {AllHtmlEntities} from 'html-entities';
+import {encode, decode} from 'html-entities';
 import replaceAll from 'string.prototype.replaceall';
 import {CONST} from './CONST';
 import {URL_REGEX} from './Url';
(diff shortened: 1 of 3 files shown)


chiragsalian:
I think there is a mistake here. I would expect a package.lock file to be uploaded also since package.json was updated. Can you verify and update it?

Also once this is merged will you also be updating the package.json file in App repo too so make use of the updated expensify-common code?

chiragsalian (approved):
LGTM

chiragsalian:
> I thought I shouldn't packageloc.json

haha all cool, every once in a while an internal engineer also forgets to update package lock. Funny enough we have a bot to check for this mistake and post in the issue but i think the bot isn't enabled for expensify-common repo. 

Either way, thanks for pushing it up 🙂

PR #546, opened 2023-06-12
fix: italic syntax removing a preceding character

eh2077:
### Fixed Issues
$ https://github.com/Expensify/App/issues/20111

# Tests
1. Go to chat and add a comment
```
h_ello_
```
2. Verify that the comment text is `h_ello_` and it's not rendered as italic style.

# QA
Same

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -181,8 +181,10 @@ export default class ExpensiMark {
                  * Use [\s\S]* instead of .* to match newline
                  */
                 name: 'italic',
-                regex: /(?!_blank")[^\W_]?_((?![\s_])[\s\S]*?[^\s_])_(?![^\W_])(?![^<]*(<\/pre>|<\/code>|<\/a>|<\/mention-user>|_blank))/g,
-                replacement: (match, g1) => (g1.includes('<pre>') || this.containsNonPairTag(g1) ? match : `<em>${g1}</em>`),
+                regex: /(\b_+|\b)(?!_blank")_((?![\s_])[\s\S]*?[^\s_])_(?![^\W_])(?![^<]*(<\/pre>|<\/code>|<\/a>|<\/mention-user>|_blank))/g,
+
+                // We add g1 which is underscores or empty string back before the <em> tag
+                replacement: (match, g1, g2) => (g2.includes('<pre>') || this.containsNonPairTag(g2) ? match : `${g1}<em>${g2}</em>`),
             },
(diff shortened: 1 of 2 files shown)


amyevans:
@mollfpr please review

lib/ExpensiMark.js
```diff
@@ -181,8 +181,11 @@ export default class ExpensiMark {
                  * Use [\s\S]* instead of .* to match newline
                  */
                 name: 'italic',
-                regex: /(?!_blank")[^\W_]?_((?![\s_])[\s\S]*?[^\s_])_(?![^\W_])(?![^<]*(<\/pre>|<\/code>|<\/a>|_blank))/g,
-                replacement: (match, g1) => (g1.includes('<pre>') || this.containsNonPairTag(g1) ? match : `<em>${g1}</em>`),
+                regex: /(\b_+|\b)(?!_blank")_((?![\s_])[\s\S]*?[^\s_])_(?![^\W_])(?![^<]*(<\/pre>|<\/code>|<\/a>|_blank))/g,
+
+                // g1 is empty string or includes one or more underscores, eg: `hello ___world_`
+                // We want to add g1 back before the <em> tag
+                replacement: (match, g1, g2) => (g2.includes('<pre>') || this.containsNonPairTag(g2) ? match : `${g1}<em>${g2}</em>`),
```
amyevans, line 188: Thoughts on updating these variables names to be more descriptive? I see that largely the convention in the file is to use `g1`, `g2`, etc, so I wouldn't consider changing it to be a blocker, but I do think it'd be slightly more self-documenting if we could use something like `extraUnderscores` and `textWithinUnderscores`, or at the least `group1`, `group2`. Open to opinions though.

mollfpr:
Thanks @amyevans will review it soon!

lib/ExpensiMark.js
mollfpr: `extraUnderscores` and `textWithinUnderscores` looks good to me. Similar to `codeFence` rule, we have `textWithinFences` instead g2.

marcochavezf:
Removing myself as reviewer, it's weird that PullerBear assigned me 😄

lib/ExpensiMark.js
eh2077: Yes, self-documenting parameter name makes sense. Updated it.

amyevans (dismissed):

amyevans (approved):

PR #548, opened 2023-06-13
fix: parsing underscores in the email of user mention 

eh2077:
### Fixed Issues
$ https://github.com/Expensify/App/issues/19957

# Tests
1. Go to a chat and add a comment `@_concierge_@expensify.com`
2. Verify that App doesn’t crash and `_concierge_` is not rendered as italic 

# QA
Same as test

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -890,6 +890,12 @@ test('Test for user mention without space or supported styling character', () =>
     expect(parser.replace(testString)).toBe(resultString);
 });
 
+test('Test for user mention with user email includes underscores', () => {
+    const testString = '@_concierge_@expensify.com';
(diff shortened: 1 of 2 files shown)


eh2077:
> @eh2077 Can we add this for other cases such as bold text?

It's a good point. I also dug this but I found that this issue happens only on underscore `_` because underscore `_` is treated as word character in JavaScript regex but `*` and `~` are not word characters. 

`*` and `~` are not valid characters for the email of user mention. For example, input like `@*concierge*@expensify.com` won’t be parsed as user mention because `*concierge*@expensify.com` is not a valid email.

So, I think we don't need to change the regex for `bold` and `italic` rule.

dangrous:
I'm happy to review this but it looks like it should be @johnmlee101 based on the linked issue?

johnmlee101 (approved):

PR #550, opened 2023-06-16
Fix: move autoEmail to before mentions

tienifr:
### Fixed Issues
$ https://github.com/Expensify/App/issues/20085

# Tests
1. Login to NewDot
2. Click on any chat
3. Send test@here@gmail.com
4. Send test+1@here@gmail.com
5. Verify the texts as emails are displayed
6. Click on the emails
7. Verify that it opens the mail sender
8. Send @test@gmail.com
9. Send @test+1@gmail.com
10. Verify the texts as mentions are displayed
11. Click on the mentions
12. Verify that it opens the user detail page of the user that was mentioned

# QA
1. Login to NewDot
2. Click on any chat
3. Send test@here@gmail.com
4. Send test+1@here@gmail.com
5. Verify the texts as emails are displayed
6. Click on the emails
7. Verify that it opens the mail sender
8. Send @test@gmail.com
9. Send @test+1@gmail.com
10. Verify the texts as mentions are displayed
11. Click on the mentions
12. Verify that it opens the user detail page of the user that was mentioned

### Screenshots/Videos
<details>
<summary>Web</summary>

https://github.com/Expensify/expensify-common/assets/113963320/1a014ebb-9e7e-46b8-b40b-6ad9e99062d6

</details>

<details>
<summary>Mobile Web - Chrome</summary>

https://github.com/Expensify/expensify-common/assets/113963320/4718dadc-cc2c-431c-984b-c1f4f58a63fd

</details>

<details>
<summary>Mobile Web - Safari</summary>

https://github.com/Expensify/expensify-common/assets/113963320/81874061-9c5a-4c9a-b2af-85b7b034202c

</details>

<details>
<summary>Desktop</summary>

https://github.com/Expensify/expensify-common/assets/113963320/da786581-1a6b-4520-affc-950f09212fd9

</details>

<details>
<summary>iOS</summary>

https://github.com/Expensify/expensify-common/assets/113963320/987db467-6954-422f-b08d-a389cefb0eb6

</details>

<details>
<summary>Android</summary>

https://github.com/Expensify/expensify-common/assets/113963320/6a6bcd4d-ba64-4c86-a446-f1ef4bc6e13a

</details>

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1049,26 +1049,62 @@ test('Skip rendering invalid markdown',() => {
     expect(parser.replace(testString)).toBe('<blockquote>*This is multiline</blockquote>bold text*');
 });
 
-test('Test for user mention with test+1@gmail.com@gmail.com', () => {
+test('Test for email with test+1@gmail.com@gmail.com', () => {
     const testString = 'test+1@gmail.com@gmail.com';
     const resultString = '<a href=\"mailto:test+1@gmail.com\">test+1@gmail.com</a>@gmail.com';
     expect(parser.replace(testString)).toBe(resultString);
 });
 
-test('Test for user mention with test@gmail.com@gmail.com', () => {
+test('Test for email with test@gmail.com@gmail.com', () => {
     const testString = 'test@gmail.com@gmail.com';
     const resultString = '<a href=\"mailto:test@gmail.com\">test@gmail.com</a>@gmail.com';
     expect(parser.replace(testString)).toBe(resultString);
 });
 
-test('Test for user mention with test@here@gmail.com', () => {
+test('Test for email with test@here@gmail.com', () => {
     const testString = 'test@here@gmail.com';
     const resultString = 'test@<a href=\"mailto:here@gmail.com\">here@gmail.com</a>';
     expect(parser.replace(testString)).toBe(resultString);
 });
 
-test('Test for user mention with test+1@here@gmail.com', () => {
+test('Test for email with test+1@here@gmail.com', () => {
     const testString = 'test+1@here@gmail.com';
     const resultString = 'test+1@<a href=\"mailto:here@gmail.com\">here@gmail.com</a>';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test user mention with @here@here.com', () => {
+    const testString = '@here@here.com';
+    const resultString = '<mention-user>@here@here.com</mention-user>';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test user mention with @abc_@here.com', () => {
+    const testString = '@abc_@here.com';
+    const resultString = '<mention-user>@abc_@here.com</mention-user>';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test user mention with @here_@here.com', () => {
+    const testString = '@here_@here.com';
+    const resultString = '<mention-user>@here_@here.com</mention-user>';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test here mention with @here_@here_.com', () => {
+    const testString = '@here_@here_.com';
+    const resultString = '<mention-here>@here</mention-here><em><mention-here>@here</mention-here></em>.com';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test here mention with @here@', () => {
+    const testString = '@here@';
+    const resultString = '<mention-here>@here</mention-here>@';
+    expect(parser.replace(testString)).toBe(resultString);
+});
+
+test('Test here mention with @here@here', () => {
+    const testString = '@here@here';
+    const resultString = '<mention-here>@here</mention-here><mention-here>@here</mention-here>';
     expect(parser.replace(testString)).toBe(resultString);
 });
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -793,43 +793,43 @@ test('Test heading1 markdown replacement with line break before or after the hea
 // Valid text that should match for user mentions
 test('Test for user mention with @username@domain.com', () => {
     const testString = '@username@expensify.com';
-    const resultString = '@<a href=\"mailto:username@expensify.com\">username@expensify.com</a>';
+    const resultString = '<mention-user>@username@expensify.com</mention-user>';
     expect(parser.replace(testString)).toBe(resultString);
 });
 
 test('Test for user mention with @phoneNumber@domain.sms', () => {
     const testString = '@+19728974297@expensify.sms';
-    const resultString = '@<a href=\"mailto:+19728974297@expensify.sms\">+19728974297@expensify.sms</a>';
+    const resultString = '<mention-user>@+19728974297@expensify.sms</mention-user>';
     expect(parser.replace(testString)).toBe(resultString);
 });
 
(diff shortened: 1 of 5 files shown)


cead22:
>Verify the texts as emails are displayed

Can you explain precisely what this means for each of the cases? For instance, is this one below correct?

image

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -896,7 +896,7 @@ test('Test for user mention without space or supported styling character', () =>
 
 test('Test for user mention with user email includes underscores', () => {
     const testString = '@_concierge_@expensify.com';
-    const resultString = '<mention-user>@_concierge_@expensify.com</mention-user>';
+    const resultString = '@_<a href=\"mailto:concierge_@expensify.com\">concierge_@expensify.com</a>';
```
cead22, line 899: Is it possible the first `_` should be part of the email?
image

cead22 (changes requested):
With this change, how does one mention a user by email?

eVoloshchak:
@tienifr, could you take a look at @cead22's concern [here](https://github.com/Expensify/expensify-common/pull/550#pullrequestreview-1484454918)?
> With this change, how does one mention a user by email?


https://github.com/Expensify/expensify-common/assets/9059945/11934c5e-35dd-42b3-bab4-c91a11537753

eVoloshchak:
> Solution 2: After digging a little bit, I notice that the current [mention rule](https://github.com/Expensify/expensify-common/blob/f9985d572c772757885ff59ea5c82b1050c6b3f7/lib/ExpensiMark.js#L113) is not replacing the match with mention if the match starts with [\.a-zA-Z]. We can revert all the previous changes (keep the original menntion -> autoEmail order), and modify the hereMentionsanduserMentionsregex so that it also ignore other email valid characters (digit, special characters...). We should modify their prefix from[`.a-zA-Z]?to[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]?`

Shouldn't it be the other way around? I.e. "replace the match with mention only if it starts with `@`"?

eVoloshchak:
Sorry for the delay, for some reason there's no review request showing up for me for this PR.
 Let's proceed with solution 2
 Is there an alternative to  `[a-zA-Z0-9.!#$%&*+/=?^_{|}~-]`? Something like `\S`?

eVoloshchak:
Looks good except one test case 

image

`dfgjl+-=32@gmail.com` is invalid email (hence not displayed as mention), but is still displayed as email

eVoloshchak:
Yeah, I'll just report tt=hat on Slack, it's not caused by this PR

eVoloshchak (approved):
LGTM
cc: @cead22

cead22 (dismissed):
This looks good. @tienifr can you add more manual tests to the PR description to test (and QA)
- Some cases of mentions to make sure those are still working
- Tapping on a mention and making sure it works
- Tapping on one of those emails you're sending, and make sure the behavior is correct

If there are more edge cases that aren't covered in the automated tests that you can think of, those would be good to add as well

cead22:
Also, I noticed the iOS videos are showing an app that looks broken, without a header. I know that's not related to these changes but if you can update those so the app doesn't look broken (since we're trying to show in these videos that the app works well), and if you know how to reproduce that, can you comment [here](https://github.com/Expensify/App/issues/20994)?

eVoloshchak:
> Also, I would like to bring attention to this edge case. It's appearing in both staging and this PR, and although I think we should resolve this in another PR, it's worth mentioning and the expected behavior should be agreed upon:

>When sending test@here@gmail.com, the here@gmail.com part is displayed as email address.
When sending @here@gmail.com or @here_@gmail.com, the @here part is displayed as a mention.

My understanding is, both @here@gmail.com and @here_@gmail.com are not valid emails, so @here part being displayed as a mention is correct I think, but curious what others think

cead22:
@tienifr
- Please update tests & QA to test tapping on an email
- Please update one of the videos to show tapping on an email, and tapping on a mention works

eVoloshchak:
> I don't know if I missed something, but I think both these statements are wrong. Those are valid emails and GH markdown renders them as such

Ah, my bad, I missed that they're supposed to be rendered as mentions instead of as emails

image

This looks like the correct behavior to me

cead22:
@tienifr whatever the behavior is today, it should be the same with this change. The reason I think the test is relevant is because we're modifying some code that's involved in rendering mentions, so imo we should test that this isn't breaking what we have

cead22:
The updates in [this comment](https://github.com/Expensify/expensify-common/pull/550#issuecomment-1615481764) make sense to me 👍 

> Would the expected behavior be that the machine will open up email app so that the user could send an email to that address?

This should be the same behavior we have today. We're just testing that we're not breaking the existing behavior since we're changing some code related to how we render emails

> Also, I believe this is complicated when recording on some emulator, since we have to install additional mailing app on that.

That's fine, we don't need this in the recordings, but we should test and QA the behavior didn't break

cead22:
> * Please update one of the videos to show tapping on an email, and tapping on a mention works

Looks like this was missed again. I still think it'd be good to show this in a video, but at least it's in the QA steps

lib/str.js
```diff
@@ -943,8 +943,14 @@ const Str = {
      * @returns {bool}
      */
     isValidMention(mention) {
-        // A valid mention starts with a space, *, _, #, ', ", or @ (with no preceding characters).
-        return /[\s*~_#'"@]/g.test(mention.charAt(0));
+        // A valid mention starts with a space or @, or
+        // starts and ends with a *, _, #, ', or " (with no preceding characters).
```
cead22, line 947: Is `#` supposed to be here? I thought `#` was used for headings like so `# heading 1`, ie, with a `#` before the text but without a `#` after the text

cead22:
@eVoloshchak since there were code changes after you approved, can you please re-review? Thanks!

lib/str.js
cead22: @eVoloshchak it can "start" with `*` in the sense that the mention can be within two `*`, so it'll still technically start with an `@`, but this is saying it can start with `*` becasue it can be like `let's ask *@carlos@expensify.com*`.

This is why I asked about `#`, cause that one is only a leading character unlike `*`, `_`, `'`, `"`, and `~`.

Incidentally @tienifr, the `~` is missing from the comment  `starts and ends with a *, _, #, ', or "...`

eVoloshchak (approved):
Looks good 

Shouldn't we have the PR reviewer checklist in these repositories too? 
You still have to go through it to test the PR

lib/str.js
```diff
@@ -943,8 +943,14 @@ const Str = {
...
+        if (/[\s@]/g.test(mention.charAt(0))) {
+            return true;
+        }
```
cead22, line 950:
```suggestion
        // Mentions can start @ preceeded by a space, eg "ping @user@domain.tld"
        if (/[\s@]/g.test(mention.charAt(0))) {
            return true;
        }
         
        // Mentions can also start and end with a *, _, ~, ', or " (with no other preceding characters).
        // eg, "ping *@user@domain.tld*"
```

cead22 (changes requested):
Minor change request, but the comment was still wrong because it said it could start and end with `#` and didn't mention `~`

> // starts and ends with a *, _, #, ', or " (with no preceding characters).

cead22:
> Shouldn't we have the PR reviewer checklist in these repositories too?

@eVoloshchak that seems like a good idea, can you start a convo in #expensify-open-source?

cead22 (approved):

Issue #547, opened 2023-06-12
Create type definition files for APIs used in App


hayata-suenaga:
Part of https://github.com/Expensify/App/issues/16665 issues.

- [Design doc](https://docs.google.com/document/d/1LNcgYeYyovQ88gNjLIgBqBReLtbLAbxeAbna9QKxeAQ/edit#bookmark=id.fiznb85a98ag)
- This library is used in other Expensify repositories, and the change should not affect these repositories.

hayata-suenaga:
Review of the PR is in progress 👍

hayata-suenaga:
The PR has been merged and the definition files were created

PR #553, opened 2023-06-21
Create type definitions for expensify-common

blazejkustra:
This is part of Expensify/App#16665 issues. The plan is to write type definitions only for the files that are used in App.

### Fixed Issues
$ #547 

# Tests

- [x] - Check that type declarations (.d.ts files) are correctly typed and in sync with Javascript implementations.
- [x] - Check that types work as expected in the App 

# QA
N/A


blazejkustra:

blazejkustra:
recheck

lib/ExpensiMark.d.ts
```diff
@@ -0,0 +1,118 @@
+declare type Replacement = string | ((...args: string[]) => string);
+declare type Rule = {
+    name: string;
+    process?: (textToProcess: string, replacement: Replacement) => string;
+    regex?: RegExp;
+    replacement: Replacement;
+    pre?: (input: string) => string;
+    post?: (input: string) => string;
+};
+export default class ExpensiMark {
```
gedu, line 10: Why a `class` and not a `type` or `interface`, I think suits better for a "class" with no body functions.

Also would be good to add when to use `class` over `interface` over `type` into the TYPESCRIPT_STYLE file

blazejkustra: Since we're typing a JavaScript class, using a class in TS is the most appropriate way to model it. This way, we're explicitly defining class constructor, methods and properties.

You suggested to use type or interface. While these options are useful, I believe they have different use cases:

- `type`: type aliases, union/intersection types, or mapped/conditional types.
- `interface`: This is great for defining a contract or the shape of an object, but isn't ideal for modeling a class with constructor and methods.

Are there some benefits of using interfaces in such case? Maybe I'm not aware of some caveats @gedu

lib/ExpensiMark.d.ts
```diff
@@ -0,0 +1,118 @@
+declare type Replacement = string | ((...args: string[]) => string);
+declare type Rule = {
+    name: string;
```
hayata-suenaga, line 3: How about making this a union of literal strings? We can collect the set of possible rule names from `ExpensiMark.js`

lib/Logger.d.ts
```diff
@@ -0,0 +1,65 @@
...
+     */
+    hmmm(message: string, parameters?: Parameters): void;
+    /**
+     * Logs a message in the browser console.
+     *
+     * @param {String} message The message to log.
+     */
+    client(message: string): void;
+}
+export {};
```
hayata-suenaga, line 65: We already have a default export for the `ExpensiMark` class, which makes this file a module. I belive we don't need this `export` statement?

lib/ExpensiMark.d.ts
```diff
@@ -0,0 +1,118 @@
...
+};
+export default class ExpensiMark {
+    rules: Rule[];
+    htmlToMarkdownRules: Rule[];
+    htmlToTextRules: Rule[];
+    constructor();
+    /**
+     * Replaces markdown with html elements
+     *
+     * @param {String} text - Text to parse as markdown
```
hayata-suenaga, line 18: @blazejkustra do we need to include types in JSDoc?

hayata-suenaga: The convention seems not to include types in JSDoc comments in d.ts files. Below is a code from `d.ts` file in `lodash` from DefinitelyTyped project

```
    interface LoDashStatic {
        /**
         * Creates an array of elements split into groups the length of size. If collection can’t be split evenly, the
         * final chunk will be the remaining elements.
         *
         * @param array The array to process.
         * @param size The length of each chunk.
         * @return Returns the new array containing chunks.
         */
        chunk<T>(array: List<T> | null | undefined, size?: number): T[][];
    }
    ```

File: lib/Logger.d.ts [hunk 5]
```diff
@@ -0,0 +1,65 @@
+declare type Parameters = string | Record<string, unknown>;
```
hayata-suenaga, line 1: I like this `Record<string, unknown>` 👍

lib/Logger.d.ts
```diff
@@ -0,0 +1,65 @@
...
+declare type ServerLoggingCallback = (logger: Logger, options: Record<string, unknown>) => Promise<void>;
```
hayata-suenaga, line 2: we can do better with the type for `options`

the only time this callback is called is [this line](https://github.com/Expensify/expensify-common/blob/f9985d572c772757885ff59ea5c82b1050c6b3f7/lib/Logger.jsx#L37).

```
const promise = this.serverLoggingCallback(this, {api_setCookie: false, logPacket: JSON.stringify(linesToLog)});
```
so we can maybe type this as `options: { api_setCookie: false, logPacket: string }`?

And for the return value, the promise is apparently resolved as an object with `requestID` property in [this line](https://github.com/Expensify/expensify-common/blob/f9985d572c772757885ff59ea5c82b1050c6b3f7/lib/Logger.jsx#L41)

it's also possible that [the return value is not actually a promise but `undefined`.
](https://github.com/Expensify/expensify-common/blob/f9985d572c772757885ff59ea5c82b1050c6b3f7/lib/Logger.jsx#L38)

```
const promise = this.serverLoggingCallback(this, {api_setCookie: false, logPacket: JSON.stringify(linesToLog)});
        if (!promise) {
            return;
        }
```

so instead of `Promise<void>` we should do `Promise<{ requestID: string }> | undefined`

lib/Logger.d.ts
```diff
@@ -0,0 +1,65 @@
...
+declare type ClientLoggingCallBack = (message: string) => void;
+export default class Logger {
+    logLines: Array<Record<string, unknown>>;
```
hayata-suenaga, line 5: [this line](https://github.com/Expensify/expensify-common/blob/f9985d572c772757885ff59ea5c82b1050c6b3f7/lib/Logger.jsx#L56) gives you more information about what an item in `logLines` looks like.

```
        const length = this.logLines.push({
            message,
            parameters,
            onlyFlushWithOthers,
            timestamp: (new Date())
        });
```

This is the only place where items are added to log lines so you don't have to make properties optional.

lib/Logger.d.ts
neil-marcellini: +1, why do we need this?

lib/Url.d.ts
```diff
@@ -0,0 +1,11 @@
...
...T|BCN|BCG|BBT|BBC|BAR|AXA|AWS|ART|APP|AOL|ANZ|AIG|AFL|AEG|ADS|ACO|ABC|ABB|AAA|ZW|ZM|ZA|YT|YE|WS|WF|VU|VN|VI|VG|VE|VC|VA|UZ|UY|US|UK|UG|UA|TZ|TW|TV|TT|TR|TO|TN|TM|TL|TK|TJ|TH|TG|TF|TD|TC|SZ|SY|SX|SV|SU|ST|SS|SR|SO|SN|SM|SL|SK|SJ|SI|SH|SG|SE|SD|SC|SB|SA|RW|RU|RS|RO|RE|QA|PY|PW|PT|PS|PR|PN|PM|PL|PK|PH|PG|PF|PE|PA|OM|NZ|NU|NR|NP|NO|NL|NI|NG|NF|NE|NC|NA|MZ|MY|MX|MW|MV|MU|MT|MS|MR|MQ|MP|MO|MN|MM|ML|MK|MH|MG|ME|MD|MC|MA|LY|LV|LU|LT|LS|LR|LK|LI|LC|LB|LA|KZ|KY|KW|KR|KP|KN|KM|KI|KH|KG|KE|JP|JO|JM|JE|IT|IS|IR|IQ|IO|IN|IM|IL|IE|ID|HU|HT|HR|HN|HM|HK|GY|GW|GU|GT|GS|GR|GQ|GP|GN|GM|GL|GI|GH|GG|GF|GE|GD|GB|GA|FR|FO|FM|FK|FJ|FI|EU|ET|ES|ER|EG|EE|EC|DZ|DO|DM|DK|DJ|DE|CZ|CY|CX|CW|CV|CU|CR|CO|CN|CM|CL|CK|CI|CH|CG|CF|CD|CC|CA|BZ|BY|BW|BV|BT|BS|BR|BO|BN|BM|BJ|BI|BH|BG|BF|BE|BD|BB|BA|AZ|AX|AW|AU|AT|AS|AR|AQ|AO|AM|AL|AI|AG|AF|AE|AD|AC|SJC|RNO|LAX)(?:\\:([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])|\\b|(?=_)))(?:(?:[.,=(+$!*]|&(?:amp|quot|#x27);)?\\/(?:[-\\w$@.+!*:(),=%~]|&(?:amp|quot|#x27);)*(?:[-\\w~@:%)]|&(?:amp|quot|#x27);)|\\/)*(?:(?:\\?(?:[-\\w$@.+!*()\\/,=%{}:;\\[\\]\\|_|~]|&(?:amp|quot|#x27);)*)?|(?:#(?:[-\\w$@.+!*()[\\],=%;\\/:~]|&(?:amp|quot|#x27);)*)?)*)";
```
hayata-suenaga, line 6: we can use template string types to use other string constant declarations to create new declarations.

```
declare const SOMETHING: 'something';
declare const ANOTHER_SOMETHING: `another ${typeof SOMETHING}`;
```

in this specific case, we can do

```
delcare URL_REGEX: `((${typeof URL_WEBSITE_REGEX})${typeof URL_PATH_REGEX}
```

neil-marcellini (commented):
Looking good so far, it's easier to read this code now. I'll continue my review another day.

lib/Url.d.ts
```diff
@@ -0,0 +1,11 @@
...
...IV|HBO|GOV|GOT|GOP|GOO|GMX|GMO|GLE|GEA|GDN|GAY|GAP|GAL|FYI|FUN|FTR|FRL|FOX|FOO|FLY|FIT|FAN|EUS|ESQ|EDU|ECO|EAT|DVR|DTV|DOT|DOG|DNP|DIY|DHL|DEV|DDS|DAY|DAD|CSC|CRS|CPA|COM|CFD|CFA|CEO|CBS|CBN|CBA|CAT|CAR|CAM|CAL|CAB|BZH|BUY|BOX|BOT|BOO|BOM|BMW|BMS|BIZ|BIO|BID|BET|BCN|BCG|BBT|BBC|BAR|AXA|AWS|ART|APP|AOL|ANZ|AIG|AFL|AEG|ADS|ACO|ABC|ABB|AAA|ZW|ZM|ZA|YT|YE|WS|WF|VU|VN|VI|VG|VE|VC|VA|UZ|UY|US|UK|UG|UA|TZ|TW|TV|TT|TR|TO|TN|TM|TL|TK|TJ|TH|TG|TF|TD|TC|SZ|SY|SX|SV|SU|ST|SS|SR|SO|SN|SM|SL|SK|SJ|SI|SH|SG|SE|SD|SC|SB|SA|RW|RU|RS|RO|RE|QA|PY|PW|PT|PS|PR|PN|PM|PL|PK|PH|PG|PF|PE|PA|OM|NZ|NU|NR|NP|NO|NL|NI|NG|NF|NE|NC|NA|MZ|MY|MX|MW|MV|MU|MT|MS|MR|MQ|MP|MO|MN|MM|ML|MK|MH|MG|ME|MD|MC|MA|LY|LV|LU|LT|LS|LR|LK|LI|LC|LB|LA|KZ|KY|KW|KR|KP|KN|KM|KI|KH|KG|KE|JP|JO|JM|JE|IT|IS|IR|IQ|IO|IN|IM|IL|IE|ID|HU|HT|HR|HN|HM|HK|GY|GW|GU|GT|GS|GR|GQ|GP|GN|GM|GL|GI|GH|GG|GF|GE|GD|GB|GA|FR|FO|FM|FK|FJ|FI|EU|ET|ES|ER|EG|EE|EC|DZ|DO|DM|DK|DJ|DE|CZ|CY|CX|CW|CV|CU|CR|CO|CN|CM|CL|CK|CI|CH|CG|CF|CD|CC|CA|BZ|BY|BW|BV|BT|BS|BR|BO|BN|BM|BJ|BI|BH|BG|BF|BE|BD|BB|BA|AZ|AX|AW|AU|AT|AS|AR|AQ|AO|AM|AL|AI|AG|AF|AE|AD|AC|SJC|RNO|LAX)(?:\\:([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])|\\b|(?=_))";
```
hayata-suenaga, line 2: let's declare `TLD_REGEX` in this file. We don't have to export it.

```
declare const TLD_REGEX: '...';
declare const URL_WEBSITE_REGEX: `...${typeof URL_WEBSITE_REGEX}...`;

File: lib/Url.d.ts [hunk 10]
```diff
@@ -0,0 +1,11 @@
...
...|BCN|BCG|BBT|BBC|BAR|AXA|AWS|ART|APP|AOL|ANZ|AIG|AFL|AEG|ADS|ACO|ABC|ABB|AAA|ZW|ZM|ZA|YT|YE|WS|WF|VU|VN|VI|VG|VE|VC|VA|UZ|UY|US|UK|UG|UA|TZ|TW|TV|TT|TR|TO|TN|TM|TL|TK|TJ|TH|TG|TF|TD|TC|SZ|SY|SX|SV|SU|ST|SS|SR|SO|SN|SM|SL|SK|SJ|SI|SH|SG|SE|SD|SC|SB|SA|RW|RU|RS|RO|RE|QA|PY|PW|PT|PS|PR|PN|PM|PL|PK|PH|PG|PF|PE|PA|OM|NZ|NU|NR|NP|NO|NL|NI|NG|NF|NE|NC|NA|MZ|MY|MX|MW|MV|MU|MT|MS|MR|MQ|MP|MO|MN|MM|ML|MK|MH|MG|ME|MD|MC|MA|LY|LV|LU|LT|LS|LR|LK|LI|LC|LB|LA|KZ|KY|KW|KR|KP|KN|KM|KI|KH|KG|KE|JP|JO|JM|JE|IT|IS|IR|IQ|IO|IN|IM|IL|IE|ID|HU|HT|HR|HN|HM|HK|GY|GW|GU|GT|GS|GR|GQ|GP|GN|GM|GL|GI|GH|GG|GF|GE|GD|GB|GA|FR|FO|FM|FK|FJ|FI|EU|ET|ES|ER|EG|EE|EC|DZ|DO|DM|DK|DJ|DE|CZ|CY|CX|CW|CV|CU|CR|CO|CN|CM|CL|CK|CI|CH|CG|CF|CD|CC|CA|BZ|BY|BW|BV|BT|BS|BR|BO|BN|BM|BJ|BI|BH|BG|BF|BE|BD|BB|BA|AZ|AX|AW|AU|AT|AS|AR|AQ|AO|AM|AL|AI|AG|AF|AE|AD|AC|SJC|RNO|LAX)(?:\\:([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])|\\b|(?=_)))(?:(?:[.,=(+$!*]|&(?:amp|quot|#x27);)?\\/(?:[-\\w$@.+!*:(),=%~]|&(?:amp|quot|#x27);)*(?:[-\\w~@:%)]|&(?:amp|quot|#x27);)|\\/)*(?:(?:\\?(?:[-\\w$@.+!*()\\/,=%{}:;\\[\\]\\|_|~]|&(?:amp|quot|#x27);)*)?|(?:#(?:[-\\w$@.+!*()[\\],=%;\\/:~]|&(?:amp|quot|#x27);)*)?)*))";
```
hayata-suenaga, line 10: see [this comment](https://github.com/Expensify/expensify-common/pull/553/files#r1237420190)

lib/str.d.ts
```diff
@@ -0,0 +1,643 @@
...
+    /**
+     * Checks if something is a certain type
+     * Stolen from underscore
+     * @param  {Mixed} obj
+     * @param  {String} type one of ['Arguments', 'Function', 'String', 'Number', 'Date',
+     *                       'RegExp', 'Error', 'Symbol', 'Map', 'WeakMap', 'Set', 'WeakSet']
+     * @return {Boolean}
+     */
+    isTypeOf(
+        obj: unknown,
```
hayata-suenaga, line 428: nice usage of `unknown` 👍

lib/str.d.ts
```diff
@@ -0,0 +1,643 @@
...
+     * JS yet, so this is a good way of doing it according to
+     * https://github.com/airbnb/javascript/issues/1439#issuecomment-306297399 and doesn't get us in trouble with
+     * linting rules.
+     *
+     * @param {String} str
+     * @param {RegExp} regex
+     *
+     * @returns {Array}
+     */
+    matchAll(str: string, regex: RegExp): unknown[];
```
hayata-suenaga, line 560: Looking at the library code, you'll realize that each match will always have `input` and `index`. 
Also, depending of the number of capturing groups te passed regex has, it has `[key: number]: string` mapping. So we can specify the return type more specifically than `unknonw[]`

`...regex: RegExp): Array<{ [key: number]: string; input: string; index: number }>;`

hayata-suenaga: I just realized we can also do something like this

```
...regex: RegExp): Array<RegExpMatchArray & { input: string; index: number }>;
```

lib/str.d.ts
```diff
@@ -0,0 +1,643 @@
...
+     * if it is a function then we will call it with
+     * any additional arguments.
+     *
+     * @param {String|Function} parameter
+     * @returns {String}
+     */
+    result<R, A>(
+        parameter: string | ((...args: A[]) => R),
+        ...args: A[]
+    ): string | R;
```
hayata-suenaga, line 593: `...args: A[]` means that all arguments have to be the same type of `A`, which is not the same for the `result` method.

I suggest the following code. This uses function overload and `extends` keyword to provide more information as well as specifying that arguments can be any number of arbitrary types.

```suggestion
    result: {
    (parameter: string): string;
    <R, A extends unknown[]>(parameter: (...args: A) => R, ...args: A): R;
  }
```

lib/str.d.ts
```diff
@@ -0,0 +1,643 @@
...
+     *
+     * @param {String} text
+     * @param {String|RegExp} searchValue
+     * @param {String|Function} replaceValue
+     * @returns {String}
+     */
+    replaceAll(
+        text: string,
+        searchValue: string | RegExp,
+        replaceValue: string | ((...args: unknown[]) => unknown)
```
hayata-suenaga, line 640:
```suggestion
        replaceValue: string | ((...args: unknown[]) => string)
```

lib/ExpensiMark.d.ts
hayata-suenaga: I don't know if rules are accessed from client codes of the `ExpensiMark` class. But if so, it might be a good idea to specify what kind of string values `name` can have.

For creating union types for `Rule` as suggested by @neil-marcellini, I don't know if we need to be that specific.

lib/ExpensiMark.d.ts
fabioh8010: About the class vs. type/interface, we don't have another option here I believe. `ExpensiMark` is used as a `class` in E/App for example and must be declared in this way.

lib/ExpensiMark.d.ts
```diff
@@ -0,0 +1,118 @@
+declare type Replacement = string | ((...args: string[]) => string);
```
fabioh8010, line 1: I saw that this type is being used in `rules` array and other methods like `modifyTextForUrlLinks`, `modifyTextForEmailLinks`, `modifyTextForQuote` and `formatTextForQuote`. The `rules` array indeed accept both string and function, but the other methods only uses `replacement` as a function. We should have a separate type for them.

lib/CONST.d.ts
```diff
@@ -0,0 +1,779 @@
+/**
+ * URL of our CloudFront Instance
+ */
+export declare const g_cloudFront: "https://d2k5nsl2zxldvw.cloudfront.net";
+/**
+ * URL of our image CDN
+ */
+export declare const g_cloudFrontImg: "https://d2k5nsl2zxldvw.cloudfront.net/images/";
+export declare const CONST: {
```
fabioh8010, line 9: @blazejkustra What about using [ReadonlyDeep](https://github.com/sindresorhus/type-fest/blob/main/source/readonly-deep.d.ts) from `type-fest`? In this way we don't need to type every property as `readonly`, making the code more readable.

lib/CONST.d.ts
```diff
@@ -0,0 +1,779 @@
...
+            readonly id: "businessClosing";
+            readonly label: "Company closing, downsizing, or acquired";
+            readonly prompt: "What software are you migrating to and what led to this decision?";
+        };
+    };
+};
+/**
+ * UI Constants
+ */
+export declare const UI: {
```
fabioh8010, line 748: @blazejkustra Same here about `ReadonlyDeep`.

lib/ExpensiMark.d.ts
blazejkustra: Rules aren't accessed in the App, so I don't think we need to type them as suggested by @neil-marcellini. Changed to union of literal strings 👌

lib/Logger.d.ts
blazejkustra: It is [Typescript default behaviour](https://github.com/microsoft/TypeScript/issues/38713) when generating declaration files, that's why it is included here.

lib/ExpensiMark.d.ts
blazejkustra: I removed typings from JSDoc in `.d.ts` files

blazejkustra:
@hayata-suenaga 

> expensify-common doesn't have index.js that exports everything? (or do we have a script that generates index.js???)

That's correct, there is no such file. I don't see any script for that as well.

> I was wondering that we might need index.d.ts from which we export all types (i.e. types used in App). but I guess if the library itself doesn't have index.js file, then we don't need to do this?

We don't need this as there is no index.js.

> also, we have index.js specified as the main file in package.json. I believe libraries also usually have to specify types or typings field for the main d.ts file that is used when the main file is imported. But as we actually don't have index.js file, we don't have to specify types field?

Exactly, no need to specify it as there is no main index.js file.

lib/Logger.d.ts
blazejkustra: Agree, typed as suggested 👌

lib/Url.d.ts
blazejkustra: I generated this file with `tsc`, I don't think it is worth the manual effort to type it with template string types. Wdyt @hayata-suenaga?

lib/Url.d.ts
blazejkustra: Done!

lib/str.d.ts
blazejkustra: Nice one! Adjusted 😄

lib/CONST.d.ts
blazejkustra: I prefer not to use type-fest for declaration files in expensify-common. This file was generated with `tsc`, I don't see advantages in making manual adjustments.

lib/ExpensiMark.d.ts
blazejkustra: I agree, adjusted the code 🙆‍♂️

lib/Url.d.ts
hayata-suenaga: ah let's don't do it then

@blazejkustra by the way, did you generate all `d.ts` files in this repo with `tsc`?

lib/Logger.d.ts
hayata-suenaga: @blazejkustra did you use `tsc` to generate this type declaration file?

if so and if this is the default behavior, we can leave it there 👍

hayata-suenaga (commented):
_I think I commented this without a link to the code... here is the relevant code portion:_ 

```ts
// Url.d.ts
declare const TLD_REGEX: "XN--VERMGENSBERATUNG-PWB|XN--VERMGENSBERATER-CTB|XN--
```

@blazejkustra I'm so sorry. I think [my comment](https://github.com/Expensify/expensify-common/pull/553#discussion_r1237422811) was misleading.

I suggested declaring TDL_REGEX in `Url.d.ts` file to be used inside string interpolations for other constants.

But we [decided not to do that](https://github.com/Expensify/expensify-common/pull/553#discussion_r1238203261)

There is no usage for `TLD_REGEX` as it's not exported from this file.

Also, for the type for `tlds.js` to be used when `tlds` is imported, we need to create a separate file `tlds.d.ts`

so your original code that puts declaration for `tlds` inside `tlds.d.ts` was better

could you revert to that again I'm sorry for the confusion 🙇 🙇 🙇

hayata-suenaga:
@blazejkustra thank you for the great job and quick fix 🙌

Just a small adjustment left: I realized that types are still left in JSDoc comments in `CONST.d.ts`

Is it possible to remove them? 🙇 

also @mountiny suggested here to add a new line at the end of each file. Unless it's a conversion not to include a new line in `d.ts` files, let's add new lines for the sake of consistency with other files 👍

roryabraham:
This PR seems pretty crazy... I was on-board with approach only as a temporary measure, but seeing the actual PR is quickly dissuading me again

1. It seems like creating and maintaining this definition file is as much or more work than just re-writing this code using TypeScript and publishing the package as JS
1. It's also much less valuable since we have [`skipLibCheck` set to `true` in our E/App tsconfig](https://github.com/Expensify/App/pull/20179/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R31), so errors in these type declaration files will be ignored.

What would be the consequence if we just didn't do this at all right now, then properly implemented TypeScript in expensify-common at a slower pace?

lib/Url.d.ts
blazejkustra: Yes, I migrated these files to Typescript and compiled to Javascript which generated declaration files alongside. I think this is the easiest and most efficient way to type it correctly 😄

blazejkustra:
> This PR seems pretty crazy... I was on-board with approach only as a temporary measure, but seeing the actual PR is quickly dissuading me again

The approach to write type declarations is only a temporary solution, `expensify-common` will be migrated to Typescript right after App migration is finished. 

> It seems like creating and maintaining this definition file is as much or more work than just re-writing this code using TypeScript and publishing the package as JS

Yes this brings more work, but it is a temporary solution and eventually declaration files will be replaced with TS implementations.

These declaration files (`.d.ts`) were generated with `tsc` - meaning that I have `.ts` files which are almost ready (there are some TS warning/errors that would have to be fixed). The problem is that `expensify-common` is quite outdated and  Typescript setup in this repo would require much more changes like removal of grunt, babel, CI/CD changes and more. We agreed that E/App is the priority and type declarations in this repo will be eventually replaced with Typescript implementations. That's why I believe it is better to just stick to type declarations **for now**.

> It's also much less valuable since we have [skipLibCheck set to true in our E/App tsconfig](https://github.com/Expensify/App/pull/20179/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R31), so errors in these type declaration files will be ignored.

Even with `skipLibCheck: false` Typescript doesn't check type declaration files with it's Javascript implementations. This flag is more about conflicts between different packages.

> What would be the consequence if we just didn't do this at all right now, then properly implemented TypeScript in expensify-common at a slower pace?

A lot of @ts-ignore/@ts-expect-error annotations in E/app, and possibly some runtime errors that comes from wrong function usage. Type declarations are much better than nothing in my opinion.

cc @hayata-suenaga @roryabraham

hayata-suenaga:
@roryabraham thank you for sharing your concern, and I apologize if we haven't communicated to you well the reasoning behind why we agreed upon creating declaration files for `expensify-common`.

> Type declarations are much better than nothing in my opinion.

I agree with @blazejkustra on this. There is no option to start migrating NewDot without doing anything to `expensify-common`. Anything is better than no type information.

So there are only two options:

1. Halt NewDot migration until we perfect typings of `expensify-common` and `onyx`
2. Create type declaration files for exposed functions/methods and push forward NewDot TS migration.

An important point to keep in mind is that type declaration files don't compromise type safety.

What do you think @roryabraham ❓

hayata-suenaga:
@roryabraham thank you for sharing your concern, and I apologize if we haven't communicated to you well the reasoning behind why we agreed upon creating declaration files for `expensify-common`.

> Type declarations are much better than nothing in my opinion.

I agree with @blazejkustra on this. There is no option to start migrating NewDot without doing anything to `expensify-common`. Anything is better than no type information.

So there are only two options:

1. Halt NewDot migration until we perfect typings of `expensify-common` and `onyx`
2. Create type declaration files for exposed functions/methods and push forward NewDot TS migration.

An important point to keep in mind is that type declaration files don't compromise type safety.

What do you think @roryabraham ❓

hayata-suenaga (dismissed):
didn't realize there were type annotations left in JSDoc in `sr.d.ts` @blazejkustra thank you for removing them 🙇 

The code looks 👍 to me now.

@blazejkustra So you copied contents of `.js` files to `.ts` files and added types so that `tsc` can create `d.ts` files? Nice idea and thank you for your work 🙌

blazejkustra:
> @blazejkustra So you copied contents of .js files to .ts files and added types so that tsc can create d.ts files? Nice idea and thank you for your work 🙌

That's exactly what I did, maybe not the fastest way but it works 😄

lib/str.d.ts
```diff
@@ -0,0 +1,585 @@
+declare const Str: {
+    /**
+     * Return true if the string is ending with the provided suffix
+     *
+     * @param str String ot search in
+     * @param suffix What to look for
+     */
+    endsWith(str: string, suffix: string): boolean;
+    /**
```
neil-marcellini, line 9: **NAB:style:** can we please have a line break between each function declaration and the comment for the next function?

lib/str.js
neil-marcellini, line 1: NAB: It would be good to write `{String|undefined}` for `getExtension` too.

neil-marcellini (changes requested):
Looking great. I think we should remove the @param annotations completely and then I'm happy with it. We still have them in JS if someone wants to read a description of the parameters.

lib/str.d.ts
```diff
@@ -0,0 +1,585 @@
+declare const Str: {
+    /**
+     * Return true if the string is ending with the provided suffix
+     *
+     * @param str String ot search in
+     * @param suffix What to look for
```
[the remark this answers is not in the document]
hayata-suenaga: I believe we need JS comments on `d.ts` files to get Intellisense hints

I don't think this is an invalid JSDoc syntax

@blazejkustra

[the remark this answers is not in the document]
blazejkustra: We decided to remove only type information from JSDoc. It's valid JSDoc that gives us more intellisense and context on each function. I'll leave it as is 👍

lib/str.d.ts
blazejkustra: I generated these files with `tsc` as described in other comment. I prefer to leave it as is, it's a lot of manual work that isn't worth IMO.

lib/str.js
blazejkustra: Agree! Adjusted @neil-marcellini

neil-marcellini (approved):

hayata-suenaga (approved):
nice work 🎉

fabioh8010 (approved):
LGTM!

AndrewGable (approved):

neil-marcellini:
I'm going to merge this because 4 of us agree it's good to go and I want to keep things moving. We can always do minor fixes in follow ups if needed.

PR #559, opened 2023-07-21
Fix: error from executing url regex

eh2077:
### Fixed Issues
$ https://github.com/Expensify/App/issues/21266

# Tests
1. Open iOS App and go to a chat
2. Add following comment
```
www.asjjssjdjdjdjdjdjjeiwiwiwowkdjdjdieikdjfidekjcjdkekejdcjdkeekcjcdidjjcdkekdiccjdkejdjcjxisdjjdkedncicdjejejcckdsijcjdsodjcicdkejdicdjejajasjjssjdjdjdjdjdjjeiwiwiwowkdjdjdieikdjfisjksksjsjssskssjskskssksksksksskdkddkddkdksskskdkdkdksskskskdkdkdkdkekeekdkddenejeodxkdndekkdjddkeemdjxkdenendkdjddekjcjdkekejdcjdkeekcjcdidjjcdkekdiccjdkejdjcjxisdjjdkedncicdjejejcckdsijcjdsodjcicdkejdicdjejajasjjssjdjdjdjdjdjjeiwiwiwowkdjdjdieikdjfidekjcjdkekejdcjdkeekcjcdidjjcdkekdiccjdkejdjcjxisdjjdkedncicdjejejcckdsijcjdsodjcicdkejdi.com
```
3. Verify App doesn’t crash after adding the comment
4. Click to edit the link and replace the link to
```
www.asjjssjdjdjdjdjdjjeiwiwiwowkdjdjdieikdjfidekjcjdkekejdcjdkeekcjcdidjjcdkekdiccjdkejdjcjxisdjjdkedncicdjejejcckdsijcjdsodjcicdkejdicdjejajasjjssjdjdjdjdjdjjeiwiwiwowkdjdjdieikdjfisjksksjsjssskssjskskssksksksksskdkddkddkdksskskdkdkdksskskskdkdkdkdkekeekdkddenejeodxkdndekkdjddkeemdjxkdenendkdjddekjcjdkekejdcjdkeekcjcdidjjcdkekdiccjdkejdjcjxisdjjdkedncicdjejejcckdsijcjdsodjcicdkejdicdjejajasjjssjdjdjdjdjdjjeiwiwiwowkdjdjdieikdjfidekjcjdkekejdcjdkeekcjcdidjjcdkekdiccjdkejdjcjxisdjjdkedncicdjejejcckdsijcjdsodjcicdkejdi.org
```
5. Verify the link is changed and the App doesn’t crash

# QA
Same as test

### DIFF
--- a/lib/ExpensiMark.d.ts
+++ b/lib/ExpensiMark.d.ts
@@ -86,7 +86,7 @@ export default class ExpensiMark {
      * @param textToCheck - Text to check
      */
     containsNonPairTag(textToCheck: string): boolean;
-    extractLinksInMarkdownComment(comment: string): string[];
+    extractLinksInMarkdownComment(comment: string): string[] | undefined;
     /**
      * Compares two markdown comments and returns a list of the links removed in a new comment.
      *
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -413,6 +413,7 @@ export default class ExpensiMark {
             });
         } catch (e) {
             Log.warn('Exception occurs in method ExpensiMark.replace', {error: e});
+
             // We want to return text without applying rules if exception occurs during replacing
             return shouldEscapeText ? _.escape(text) : text;
         }
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -1,5 +1,4 @@
 import _ from 'underscore';
-import Log from './Log';
 import Str from './str';
 import {MARKDOWN_URL_REGEX, LOOSE_URL_REGEX, URL_REGEX} from './Url';
 import {CONST} from './CONST';
@@ -412,7 +411,8 @@ export default class ExpensiMark {
                 }
             });
         } catch (e) {
-            Log.warn('Error replacing text with html in ExpensiMark.replace', {error: e});
+            // eslint-disable-next-line no-console
+            console.warn('Error replacing text with html in ExpensiMark.replace', {error: e});
 
             // We want to return text without applying rules if exception occurs during replacing
             return shouldEscapeText ? _.escape(text) : text;
@@ -751,7 +751,8 @@ export default class ExpensiMark {
             const links = _.map(matches, match => Str.sanitizeURL(match[2]));
             return links;
         } catch (e) {
-            Log.warn('Error parsing url in ExpensiMark.extractLinksInMarkdownComment', {error: e});
+            // eslint-disable-next-line no-console
+            console.warn('Error parsing url in ExpensiMark.extractLinksInMarkdownComment', {error: e});
             return undefined;
         }
     }
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -412,7 +412,7 @@ export default class ExpensiMark {
                 }
             });
         } catch (e) {
-            Log.warn('Exception occurs in method ExpensiMark.replace', {error: e});
+            Log.warn('Error replacing text with html in ExpensiMark.replace', {error: e});
 
             // We want to return text without applying rules if exception occurs during replacing
             return shouldEscapeText ? _.escape(text) : text;
@@ -751,7 +751,7 @@ export default class ExpensiMark {
             const links = _.map(matches, match => Str.sanitizeURL(match[2]));
             return links;
         } catch (e) {
-            Log.warn('Exception occurs in method ExpensiMark.extractLinksInMarkdownComment', {error: e});
+            Log.warn('Error parsing url in ExpensiMark.extractLinksInMarkdownComment', {error: e});
             return undefined;
         }
     }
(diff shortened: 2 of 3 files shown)


eh2077:
@allroundexperts Should I move methods [Url.getURLObject](https://github.com/Expensify/App/blob/411d466cabe13b67de1d5da5a61aef303ef38a1f/src/libs/Url.js#L20) and [ValidationUtils.isValidWebsite](https://github.com/Expensify/App/blob/411d466cabe13b67de1d5da5a61aef303ef38a1f/src/libs/ValidationUtils.js#L239-L240) from App to this repo? Or just add try/catch to handle them in the App repo?

eh2077:
> @eh2077 Can you please add tests for this?

@allroundexperts Added test

lib/ExpensiMark.d.ts
```diff
@@ -86,7 +86,7 @@ export default class ExpensiMark {
      * @param textToCheck - Text to check
      */
     containsNonPairTag(textToCheck: string): boolean;
-    extractLinksInMarkdownComment(comment: string): string[];
+    extractLinksInMarkdownComment(comment: string): string[] | undefined;
```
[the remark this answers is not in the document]
eh2077: @allroundexperts I tend to stick with return `undefined` when exception occurs for the following reason. This method currently is only used by method `getRemovedMarkdownLinks` which needs to call this method twice to do diff to get the removed links. My point here is to let method `getRemovedMarkdownLinks` return empty array `[]` if any of the two links array is `undefined`, which means don't remove links if any exception occurs. Does it make sense to you?

Let's say, there's an old comment with multiple links but one of them causes regex execution error, in this case, we return `undefined` to skip removing any links. If we return `[]` for the old comment, then method `getRemovedMarkdownLinks` will remove all links.

```
Old comment
asjjssjdjdjdjdjdjjeiwiwiwowkdjdjdieikdjfidekjcjdkekejdcjdkeekcjcdidjjcdkekdiccjdkejdjcjxisdjjdkedncicdjejejcckdsijcjdsodjcicdkejdicdjejajasjjssjdjdjdjdjdjjeiwiwiwowkdjdjdieikdjfisjksksjsjssskssjskskssksksksksskdkddkddkdksskskdkdkdksskskskdkdkdkdkekeekdkddenejeodxkdndekkdjddkeemdjxkdenendkdjddekjcjdkekejdcjdkeekcjcdidjjcdkekdiccjdkejdjcjxisdjjdkedncicdjejejcckdsijcjdsodjcicdkejdicdjejajasjjssjdjdjdjdjdjjeiwiwiwowkdjdjdieikdjfidekjcjdkekejdcjdkeekcjcdidjjcdkekdiccjdkejdjcjxisdjjdkedncicdjejejcckdsijcjdsodjcicdkejdi.cdjd
www.google.com
www.expensify.com
```

```
New comment
www.google.com
www.expensify.com
```

lib/ExpensiMark.js
```diff
@@ -731,15 +740,20 @@ export default class ExpensiMark {
...
-        return links;
+        try {
+            const escapedComment = _.escape(comment);
+            const matches = [...escapedComment.matchAll(MARKDOWN_LINK_REGEX)];
+
+            // Element 1 from match is the regex group if it exists which contains the link URLs
+            const links = _.map(matches, match => Str.sanitizeURL(match[2]));
+            return links;
+        } catch (e) {
+            Log.warn('Exception occurs in method ExpensiMark.extractLinksInMarkdownComment', {error: e});
```
joelbettner, line 754: Let's update this to be a bit more specific...

```
Log.warn('Error parsing url in ExpensiMark.extractLinksInMarkdownComment', {error: e});
```

lib/ExpensiMark.js
```diff
@@ -391,23 +392,31 @@ export default class ExpensiMark {
...
+                    replacedText = replacedText.replace(rule.regex, rule.replacement);
+                }
+
+                // Post-process text after applying regex
+                if (rule.post) {
+                    replacedText = rule.post(replacedText);
+                }
+            });
+        } catch (e) {
+            Log.warn('Exception occurs in method ExpensiMark.replace', {error: e});
```
joelbettner, line 415: Let's make this:

```
Log.warn('Error replacing text with html in ExpensiMark.replace', {error: e});
```

joelbettner (changes requested):
Only requesting a couple small changes to the error messages.

lib/ExpensiMark.js

lib/ExpensiMark.js

eh2077:
@allroundexperts I found an error when sending log to backend using Logger setting from this repo https://github.com/Expensify/expensify-common/blob/main/lib/Log.jsx

It looks like that the App doesn't include the jQuery in the build.



The logger from App repo seems working well https://github.com/Expensify/App/blob/main/src/libs/Log.js

Any suggestions on how to fix this issue?

cc @joelbettner

eh2077:
@allroundexperts I dug the logging error stack and I feel it's a bit tricky. I also found that jQuery may not be able to use in native platform. So I think we can use `console.warn` to avoid using the logger.

joelbettner (approved):

PR #561, opened 2023-07-28
Handle currency symbols with a period in fromUSDToNumber

Li357:
### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/303703

# Tests
Added a unit test for Bs.S, the Venezuelan currency that has a period in the symbol itself. Before, this was parsing as NaN.

# QA
In OldDot, Verify that when changing Settings > Policies > Reports > Report Currency to Bs.S then going to Settings > Policies > Expenses > Time and setting the Default Hourly Rate to Bs.S6.00 correctly autosaves.

### DIFF
--- a/__tests__/Str-test.js
+++ b/__tests__/Str-test.js
@@ -95,3 +95,9 @@ describe('Str.sanitizeURL', () => {
         expect(Str.sanitizeURL('HTtp://FOO.com/blah_BLAH')).toBe('http://foo.com/blah_BLAH');
     });
 });
+
+describe('Str.fromUSDToNumber', () => {
+    it('Handles currency symbols with a period', () => {
+        expect(Str.fromUSDToNumber('Bs.S2.48')).toBe('248');
+    });
+});
--- a/__tests__/Str-test.js
+++ b/__tests__/Str-test.js
@@ -96,8 +96,34 @@ describe('Str.sanitizeURL', () => {
     });
 });
 
-describe('Str.fromUSDToNumber', () => {
+describe('Str.fromCurrencyToNumber', () => {
+    it('Handles negative amounts with minus sign', () => {
+        expect(Str.fromCurrencyToNumber('-$5.23')).toBe(-523);
+        expect(Str.fromCurrencyToNumber('$-5.23')).toBe(-523);
+    });
+
+    it('Handles negative amounts with ()', () => {
+        expect(Str.fromCurrencyToNumber('($5.23)')).toBe(-523);
+    });
+
+    it('Handles fractional cents when allowed', () => {
+        expect(Str.fromCurrencyToNumber('$5.223', true)).toBe(522.3);
+    });
+
+    it('Handles amounts without leading zeros', () => {
+        expect(Str.fromCurrencyToNumber('$.23')).toBe(23);
+    });
+
+    it('Handles amounts without cents', () => {
+        expect(Str.fromCurrencyToNumber('$5')).toBe(500);
+    });
+
     it('Handles currency symbols with a period', () => {
-        expect(Str.fromUSDToNumber('Bs.S2.48')).toBe('248');
+        expect(Str.fromCurrencyToNumber('Bs.S2.48')).toBe(248);
+        expect(Str.fromCurrencyToNumber('Bs.S-2.48')).toBe(-248);
+        expect(Str.fromCurrencyToNumber('-Bs.S2.48')).toBe(-248);
+        expect(Str.fromCurrencyToNumber('(Bs.S2.48)')).toBe(-248);
+        expect(Str.fromCurrencyToNumber('Bs.S.48')).toBe(48);
+        expect(Str.fromCurrencyToNumber('Bs.S2')).toBe(200);
     });
 });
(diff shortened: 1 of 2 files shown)


lib/str.js
```diff
@@ -24,16 +24,19 @@ const Str = {
...
      *
-     * @param {String}  amountStr     A string representing a USD value.
+     * @param {String}  amountStr     A string representing a currency symbol and value, like $4.02 or Bs.S97.9
      * @param {Boolean} allowFraction Flag indicating if fractions of cents should be
-     *                               allowed in the output.
+     *                                allowed in the output.
      *
      * @return {Number} The cent value of the @p amountStr.
      */
     fromUSDToNumber(amountStr, allowFraction) {
```
Li357, line 35: Before we merge, we should consider if we want to rename this method since it now does double-duty by handling all types of currency strings (symbols + values) instead of just USD.

Li357:

__tests__/Str-test.js
```diff
@@ -95,3 +95,9 @@ describe('Str.sanitizeURL', () => {
         expect(Str.sanitizeURL('HTtp://FOO.com/blah_BLAH')).toBe('http://foo.com/blah_BLAH');
     });
 });
+
+describe('Str.fromUSDToNumber', () => {
+    it('Handles currency symbols with a period', () => {
+        expect(Str.fromUSDToNumber('Bs.S2.48')).toBe('248');
```
dangrous, line 101:
```suggestion
        expect(Str.fromUSDToNumber('Bs.S2.48')).toBe(248);
```
Returns a number not a string, looks like

lib/str.js
```diff
@@ -24,16 +24,19 @@ const Str = {
...
 
     /**
-     * Converts a USD string into th number of cents it represents.
+     * Converts a currency string into the number of cents it represents.
      *
-     * @param {String}  amountStr     A string representing a USD value.
+     * @param {String}  amountStr     A string representing a currency symbol and value, like $4.02 or Bs.S97.9
      * @param {Boolean} allowFraction Flag indicating if fractions of cents should be
-     *                               allowed in the output.
+     *                                allowed in the output.
```
dangrous, line 31: given that the line above is longer now I think we can consolidate this to one line.

lib/str.js
```diff
@@ -24,16 +24,19 @@ const Str = {
...
      *
      * @return {Number} The cent value of the @p amountStr.
      */
     fromUSDToNumber(amountStr, allowFraction) {
-        let amount = String(amountStr).replace(/[^\d.\-()]+/g, '');
+        amountStr = String(amountStr);
+        // Currently, no currency symbol has a number in it, so we find the first number
+        let withoutCurrency = amountStr.substring(amountStr.search(/\d/));
```
dangrous, line 38: Will amounts always have a leading zero? like would we ever pass `$.25` or something like that? If so this might return the wrong value. (I'm not sure on the answer to this)

lib/str.js
dangrous: Yeah I would. `fromCurrencyToNumber` maybe? Just need to make sure we replace it everywhere

dangrous (changes requested):
Couple comments and questions! GH checks are also a bit sad.

Li357:
Good catches, thanks! Turns out this was a little more annoying than I though. I've changed the regex to match based on the end of the amount string to match either `\d+\.?\d*` for `5` or `5.233` or `5.`, or `\.\d+` for no-leading-zero amounts. I've added more tests for success cases that match the current behavior. There are some cases that now are parsed with this new code, such as `($5.00` returns 500 instead of NaN like before but I don't see that as too important.

dangrous (approved):
Looks good!

lib/str.js
```diff
@@ -24,19 +24,25 @@ const Str = {
...
+     * @param {Boolean} allowFraction Flag indicating if fractions of cents should be allowed in the output.
      *
      * @return {Number} The cent value of the @p amountStr.
      */
-    fromUSDToNumber(amountStr, allowFraction) {
-        let amount = String(amountStr).replace(/[^\d.\-()]+/g, '');
-        if (amount.match(/\(.*\)/)) {
-            const modifiedAmount = amount.replace(/[()]/g, '');
-            amount = `-${modifiedAmount}`;
+    fromCurrencyToNumber(amountStr, allowFraction) {
```
techievivek, line 34: It looks like changes introduced in this PR have bough this regression https://github.com/Expensify/Expensify/issues/309211

1. If the amount field is not present in the CSV file, it throws an error, originally we used to default to 0.
2. If the amount is negative the amount calculated is `NaN` and further defaulting to 0?

Screenshot 2023-08-21 at 4 41 23 PM
Screenshot 2023-08-21 at 4 42 56 PM
Screenshot 2023-08-21 at 4 58 26 PM
Screenshot 2023-08-21 at 4 58 46 PM

techievivek: This almost seems to cause another regression here: https://github.com/Expensify/Expensify/issues/309227

dangrous: Can you have a negative per diem? Seems like defaulting to 0 in that case would be correct, right?

For the other, maybe something like changing [this line](https://github.com/Expensify/expensify-common/pull/561/files?diff=unified#diff-eb3aad4f6d64ed33948e40b58a65d583ddf513521d7a3e23871e5f275ac907fcR47) to have a default case of 0?

Will defer to @Li357 though

PR #562, opened 2023-07-31
Update email regex

cristipaval:
We don't want the front end to pass emails that are not valid in the backend. We use [FILTER_VALIDATE_EMAIL](https://www.php.net/manual/en/filter.constants.php) from php which validates the emails if they match RFC-822 standards. According to those standards, the username part of the email can't be longer than 64 characters. This PR adds this condition in our regex for the email validation in JS.

See [here](https://expensify.slack.com/archives/C03TQ48KC/p1690482472889119) a discussion about the email validation inconsistency throughout the project.

### Fixed Issues
Part of https://github.com/Expensify/App/issues/23180

# Tests
1. In NewDot, go to **Profile** -> **Contact Method** -> **New Contact Method**
2. In the email input field add an email with a long username, like `usernamelongerthan64charactersshouldnotworkaccordingtorfc822whichisusedbyphp@gmail.com`
3. Click on **Add** button
4. Verify that you get an **Invalid contact method** error

# QA
We'll be QA'd with the App PR

### DIFF
--- a/__tests__/Str-test.js
+++ b/__tests__/Str-test.js
@@ -95,3 +95,11 @@ describe('Str.sanitizeURL', () => {
         expect(Str.sanitizeURL('HTtp://FOO.com/blah_BLAH')).toBe('http://foo.com/blah_BLAH');
     });
 });
+
+describe('Str.isValidEmail', () => {
+    it('Correctly detects a valid email', () => {
+        expect(Str.isValidEmail('abc@gmail.com')).toBeTruthy();
+        expect(Str.isValidEmail('test@gmail')).toBeFalsy();
+        expect(Str.isValidEmail('usernamelongerthan64charactersshouldnotworkaccordingtorfc822whichisusedbyphp@gmail.com')).toBeFalsy();
+    });
+});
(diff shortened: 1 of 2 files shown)


__tests__/Str-test.js
```diff
@@ -95,3 +95,11 @@ describe('Str.sanitizeURL', () => {
         expect(Str.sanitizeURL('HTtp://FOO.com/blah_BLAH')).toBe('http://foo.com/blah_BLAH');
     });
 });
+
+describe('Str.isValidEmail', () => {
+    it('Correctly detects a valid email', () => {
+        expect(Str.isValidEmail('abc@gmail.com')).toBeTruthy();
+        expect(Str.isValidEmail('test@gmail')).toBeFalsy();
+        expect(Str.isValidEmail('usernamelongerthan64charactersshouldnotworkaccordingtorfc822whichisusedbyphp@gmail.com')).toBeFalsy();
```
deetergp, line 103: NAB: Should we add a test for usernames shorter than 1 character?

deetergp (approved):
Code looks good. I had one minor NAB comment.

__tests__/Str-test.js
cristipaval: Ah yes, it makes sense to add that test for completeness. Thanks 🙇 
Follow-up PR [here](https://github.com/Expensify/expensify-common/pull/563)

PR #567, opened 2023-08-21
Handle cases without currency symbols

Li357:
### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/309211
$ https://github.com/Expensify/Expensify/issues/309227

# Tests
Added two unit tests to test cases without currency symbols and without passing a currency string at all.

# QA

Run the QA steps in the linked issues and verify that the error does not occur.

### DIFF
--- a/__tests__/Str-test.js
+++ b/__tests__/Str-test.js
@@ -126,8 +126,19 @@ describe('Str.fromCurrencyToNumber', () => {
         expect(Str.fromCurrencyToNumber('Bs.S.48')).toBe(48);
         expect(Str.fromCurrencyToNumber('Bs.S2')).toBe(200);
     });
+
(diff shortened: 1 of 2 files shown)


techievivek:
@Li357 QQ: Did you confirm if those 2 issues are no longer reproducible?

Li357:
Just confirmed https://github.com/Expensify/Expensify/issues/309211 is no longer producible

Li357:
@techievivek Verified both are not reproducible!

techievivek:
I confirmed the other one is not reproducible with this change.

techievivek (approved):
Looks good. Thanks for the quick fix.

techievivek:
@Li357 Let's bump the new version in Web-E and I can CP that to staging.

PR #575, opened 2023-09-13
Removing special chars that are in the end of a url and shouldn't be

ShogunFire:
Because we are removing unmatched end parenthesis, some links that end with special characters are accepted when they shouldn't be. we need to remove those special characters.

### Fixed Issues
$ https://github.com/Expensify/App/issues/25532

# Tests
1. Go to any repórt
2. Send a link with a slash for example: google.com/toto
3. Edit the message and add one of those characters to the link in the parenthese: `$*.+!,=`
4. When you validate the edit verify that the text is white without links like this:
image

5. Send another message google.com/toto.) or google.com/toto$) for example and verify that the special char before the parenthese is not blue (not included in the link)
image

# QA
Same as above and test links in general

Video for web:

https://github.com/Expensify/expensify-common/assets/19537677/8bd1d2ca-5695-412a-8690-448a46295b4c

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -470,7 +470,7 @@ export default class ExpensiMark {
             }
 
             // Because we are removing ) parenthesis, some special characters that shouldn't be in the href are in the href
-            // For example google.com/toto.) is accepted by the regexp, above we remove the ) parenthese so the link becomes google.com/toto. which is not a valid link
+            // For example google.com/toto.) is accepted by the regular expression above and we remove the ) parenthesis, so the link becomes google.com/toto. which is not a valid link
             // In that case we should also remove the "."
             let numberOfCharsToRemove = 0;
 
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -482,8 +482,8 @@ export default class ExpensiMark {
                     break;
             }
             if (numberOfCharsToRemove){
-                match[0] = match[0].substr(0, match[0].length - numberOfCharsToRemove);
-                match[2] = match[2].substr(0, match[2].length - numberOfCharsToRemove);
+                match[0] = match[0].substring(0, match[0].length - numberOfCharsToRemove);
+                match[2] = match[2].substring(0, match[2].length - numberOfCharsToRemove);
             }
 
             replacedText = replacedText.concat(textToCheck.substr(startIndex, (match.index - startIndex)));
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -475,7 +475,7 @@ export default class ExpensiMark {
             var numberOfCharsToRemove = 0;
 
             for (let i = match[2].length - 1; i >= 0; i--) {
-                if ("$*.+!(,=".includes(match[2][i])) {
+                if (CONST.SPECIAL_CHARS_TO_REMOVE.includes(match[2][i])) {
                     numberOfCharsToRemove++;
                 }
                 else
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -469,7 +469,7 @@ export default class ExpensiMark {
                 }
             }
 
-            // Because we are removing ) parenthese, some special characters that shouldn't be in the href are in the href
+            // Because we are removing ) parenthesis, some special characters that shouldn't be in the href are in the href
             // For example google.com/toto.) is accepted by the regexp, above we remove the ) parenthese so the link becomes google.com/toto. which is not a valid link
             // In that case we should also remove the "."
             let numberOfCharsToRemove = 0;
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -474,6 +474,24 @@ export default class ExpensiMark {
                     match[2] = match[2].substr(0, match[2].length + brace);
                 }
             }
+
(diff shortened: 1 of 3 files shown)


ShogunFire:

ShogunFire:
@0xmiroslav

ShogunFire:
Ok I am adding test cases for autolink url with special chars and unmatched parenthese in ExpensiMark-HTML-test file, I think that should be enough.

ShogunFire:
I added the video and some test cases

MonilBhavsar:
Requesting a review

lib/ExpensiMark.js
```diff
@@ -468,6 +468,24 @@ export default class ExpensiMark {
...
+                    numberOfCharsToRemove++;
+                }
+                else
+                    break;
+            }
+            if (numberOfCharsToRemove){
+                match[0] = match[0].substr(0, match[0].length - numberOfCharsToRemove);
+                match[2] = match[2].substr(0, match[2].length - numberOfCharsToRemove);
+            }
+
```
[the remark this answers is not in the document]
ShogunFire: There is probably a way to make that work in the same code but I don't think it makes sense to try to do it because the code for `)` has to count the parentheses in the left to right order while this code starts from the end.
Also this situation where we have the special chars at the end only happens after we have remove the `)`

I am adding the special characters in CONST right now

ShogunFire:
@0xmiroslav The special chars are in the const file now

ShogunFire:
@0xmiroslav I made the suggestions

lib/ExpensiMark.js
```diff
@@ -468,6 +468,24 @@ export default class ExpensiMark {
                     match[2] = match[2].substr(0, match[2].length + brace);
                 }
             }
+
+            // Because we are removing ) parenthese, some special characters that shouldn't be in the href are in the href
```
MonilBhavsar, line 472:
```suggestion
            // Because we are removing ) parenthesis, some special characters that shouldn't be in the href are in the href
```

lib/ExpensiMark.js
```diff
@@ -468,6 +468,24 @@ export default class ExpensiMark {
...
+            // For example google.com/toto.) is accepted by the regexp, above we remove the ) parenthese so the link becomes google.com/toto. which is not a valid link
```
MonilBhavsar, line 473:
```suggestion
            // For example google.com/toto.) is accepted by the regular expression above and we remove the ) parenthesis, so the link becomes google.com/toto. which is not a valid link
```

MonilBhavsar (commented):
Small comment. Looks good. Thanks for adding the tests!

MonilBhavsar:
@ShogunFire the Test Javascript is failing for some reason. Could you please merge main and check

ShogunFire:
Ok so it looks like I was wrong about those special characters never being at the end of an url. In fact those characters are never at the end of a `URL_PATH_REGEX` but they can be at the end of strings that are after the path like `URL_PARAM_REGEX` or `URL_FRAGMENT_REGEX` here: https://github.com/Expensify/expensify-common/blob/cb1c6b75cb593990ed18381ba7dba8e8ceb9f16b/lib/Url.js#L7-L9

In their tests they have `=` that is at the end and it is valid because it is part of the parameter of the url. One way to resolve this is that before I remove the special chars that are at the end I must test that we don't have a param or a fragment by testing if we have a `?` or `#` in the url, do you think this would be the right way ?

ShogunFire:
Hello @0xmiroslav Can you check this: https://github.com/Expensify/expensify-common/pull/575#issuecomment-1727814421

ShogunFire:
With the current regexp we have it's not only = that can be at the end, all those characters `$*.+!,` can also be at the end if we have a param, I don't think we should remove them either if they are part of a param.

And never removing = will create an inconsistency: `google.com/toto=)` the = will be part of the url while in `google.com/toto=` it won't be in the url

ShogunFire:
Without `?` this is not a param I think

ShogunFire:
Yes, this is what I propose:

> One way to resolve this is that before I remove the special chars that are at the end I must test that we don't have a param or a fragment by testing if we have a ? or # in the url

MonilBhavsar:
I also agree with the proposal 👍 
Just curious how we're thinking to implement it?

ShogunFire:
@0xmiroslav @MonilBhavsar I commited, I just tested like this:
` if (!url.includes('?') && !url.includes('#')) {`
is it enough ? 

I merged with main and replaced the `match[2]` with `url` since we created this variable in the other PR.

I added tests for params and fragments

MonilBhavsar:
I think so it will work. What could be the other cases though 🤔 
@0xmiroslav could you please take a look at the comment above?

ShogunFire:
@0xmiroslav This question was for me ? I don't know but it was like this in the regexp, I think the regexp might not be perfect but we can see that on different websites (github and slack for example) they have different rules so I would leave it like this just in case we are missing something

ShogunFire:
@0xmiroslav Yes I think we are good now

ShogunFire:
@0xmiroslav I pulled main

MonilBhavsar (approved):
Thanks! Sorry, I missed it somehow

MonilBhavsar:
I'm pretty sure, the tests were passing

Screenshot 2023-10-17 at 11 36 54 AM

PR #632, opened 2024-01-15
fix: remove String.replaceAll polyfill

aswin-s:
This PR removes the String.replaceAll ployfill in an attempt to reduce the bundle size. String.replaceAll method is natively supported from NodeJS v15 onwards. So the polyfill is not needed any more. This removes appoximately 30KB from overall bundle size. 

### Fixed Issues
$ GH_LINK https://github.com/Expensify/App/issues/31183

# Tests
No new test cases required. Str.replaceAll method now makes use of built-in String.replace method. So all existing test cases should pass.

     - Verified that all existing testcases pass
     - Verified all functionalities are working as expected by linking into New-Expensify

# QA

    Same as tests

     Str.replaceAll is used in Emoji and comment parsing in Expensify app. Test by adding/removing emojis in chats.

### DIFF
--- a/lib/str.js
+++ b/lib/str.js
@@ -1,7 +1,6 @@
 /* eslint-disable no-control-regex */
 import _ from 'underscore';
 import {encode, decode} from 'html-entities';
-import replaceAll from 'string.prototype.replaceall';
 import {CONST} from './CONST';
 import {URL_REGEX} from './Url';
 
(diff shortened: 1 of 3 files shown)


aswin-s:

aswin-s:
recheck

neil-marcellini (approved):
Looks good, thanks

neil-marcellini:
I also confirmed that all our repos using expensify-common are using a node version >= 15

PR #644, opened 2024-02-08
Stop escaping codeFence data-code-raw attribute data

robertKozik:
This PR is a follow up for [live markdown preview web PR](https://github.com/Expensify/App/issues/27977) as there was discovered a bug which prevent parsing when special character was inside codeFence fe. ` ``` one > two ``` `.

To fix this issue we shouldn't escape data inside codeFence `data-code-raw` attribute so the input and output of the parser will be the same for the `react-native-live-markdown` library

### Fixed Issues
$ https://github.com/Expensify/App/issues/27977

# Tests
There are already lots of unit tests present for covering this behaviour. All relevant ones were adjusted to new behaviour
I've checked it on example app of `react-native-live-markdown` library which we are using for text input inside expensify app

To test the behaviour you should use example app of the [react-native-live-markdown library](https://github.com/Expensify/react-native-live-markdown). To do so:
1. change the expensify-common version inside `/parser` directory and reinstall dependencies (`npm install`)
2. build parser output file (`npm run build`)
3. Build and open example app on desired platform (iOS/Android/Web)
4. Type in code fence with special character inside it (fe. ` ``` one > two ``` `)
5. Verify that input is parsed as codeFence properly

# QA
There should be no regressions involved as I are only changing the behaviour of html attributed used only in `react-native-live-markdown` library. Any regression are possible only after new version of `react-native-live-markdown` library is present inside E/App

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1726,7 +1726,7 @@ describe('when should keep raw input flag is enabled', () => {
 
     test('quote with other markdowns', () => {
         const quoteTestStartString = '>This is a *quote* that started on a new line.\nHere is a >quote that did not\n```\nhere is a codefenced quote\n>it should not be quoted\n```';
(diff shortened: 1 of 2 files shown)


shubham1206agra:
@robertKozik Is it possible to test this in NewDot now?
If not, I will approve this right away.

marcaaron (approved):

PR #654, opened 2024-02-19
Add support for nested quotes

robertKozik:
This PR adds support for nested quotes. Maximal level of quotes is set to *3*. 
Additionally it fixes bug with double removal of `>` symbol

### Fixed Issues
$ https://github.com/Expensify/App/issues/36227

# Tests

Every test involving writing messages with quotes will be relevant in this case, as the behaviour of single quotes shouldn't be changed. In addition, every such a test can be tweaked to check nested quotes case simply by adding additional `>` symbol
I've checked it by running all unit tests and by linking it in E/App and checking in chat itself

I'm going to add additional unit tests which should be sufficient to cover all popular cases. Morove manual testing inside E/App.

# QA
Typing any message with multiple quotes (f.e `> > > Hello world`)
Chat/Quotes

Anything regarding quotes should be tested, as quite big chunk of code regarding quotes were changed here


blimpich (approved):
LGTM

PR #658, opened 2024-02-29
Add support for inline image Markdown

kidroca:
This Pull Request introduces enhancements to the ExpensiMark library by implementing the conversion of image markdown syntax to HTML `` tags and vice versa. It covers various cases including handling single images and text containing images. 
Additionally, it addresses the conversion of `` tags back to markdown. 
An issue with image alt text containing markdown not converting to plain text is acknowledged and currently marked as a limitation.

### Fixed Issues
$ https://github.com/Expensify/App/issues/37246

# Tests
1. **Unit/Integration Tests Covering the Change:**
   - The changes are covered by unit tests added to `ExpensiMark-HTML-test.js`, `ExpensiMark-HTMLToText-test.js`, and `ExpensiMark-Markdown-test.js`. These tests validate the conversion of image markdown to HTML tags, the handling of images with different attributes, and the conversion of HTML `` tags back to markdown.
   
2. **Tests Performed to Validate Changes:**
   - Ran all newly added tests and ensured they pass, verifying the correct conversion of markdown to HTML `` tags and vice versa.
   - Manually tested the parsing of markdown containing images in various formats and contexts to ensure the output matches expectations.
   - Checked the handling of invalid image URLs to ensure they remain unchanged, as expected.

# QA
1. **QA Validation Steps:**
   - Utilize NewDot to post comments incorporating Markdown (MD) image syntax along with public images sourced from platforms like Unsplash. Verify that these images are correctly displayed inline within the comments.
   - Ensure that you can seamlessly mix text and images within the same comment, maintaining the integrity of the intended message and layout.
   - Confirm the functionality to edit previously submitted comments, checking that upon editing, the comments are accurately presented with the original Markdown syntax intact.
   - Test the handling of invalid Markdown image syntax, such as using a broken or incorrectly formatted URL (e.g., `image`). These should not be parsed and ought to remain displayed as the raw Markdown text, without conversion or rendering as an image.

2. **Areas to Test for Regressions:**
   - General markdown to HTML conversion functionality, especially focusing on links, autolinks, and other media types to ensure no regressions have been introduced.
   - The handling of alt text in both markdown and HTML conversions, given the new logic around image processing.
   - The interaction of image markdown with other markdown features (e.g., bold, italic) to verify that the integration is seamless and does not introduce unexpected behavior.

PR #661, opened 2024-03-12
Add support for inline image Markdown - attribute fix

kidroca:
This PR is an optional followup for https://github.com/Expensify/expensify-common/pull/658 if we'd like to patch a problem where image attribute content is being parsed to HTML as detailed in: https://github.com/Expensify/expensify-common/pull/658#discussion_r1508088239

### Fixed Issues
$ https://github.com/Expensify/App/issues/37246

# Tests
1. **Unit/Integration Tests Covering the Change:**
   - The changes are covered by unit tests added to `ExpensiMark-HTML-test.js` and `ExpensiMark-Markdown-test.js`. These tests validate that MD content designated for the `alt` attribute is escaped as HTML entities so it isn't converted to html tags by following rules
   
2. **Tests Performed to Validate Changes:**
   - Ran all newly added tests and ensured they pass, verifying the correct conversion of markdown to HTML `` tags and vice versa.
   - Manually tested the parsing of markdown containing images in various formats and contexts to ensure the output matches expectations.
   - Checked the handling of invalid image URLs to ensure they remain unchanged, as expected.

# QA
1. **QA Validation Steps:**
   - Test the handling of invalid Markdown image syntax, such as using a broken or incorrectly formatted URL (e.g., `image`). These should not be parsed and ought to remain displayed as the raw Markdown text, without conversion or rendering as an image.

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1931,4 +1931,10 @@ describe('Image markdown conversion to html tag', () => {
         const resultString = '<img src=\"https://example.com/image.png\" alt=\"test&quot; onerror=&quot;alert(&#x27;xss&#x27;)\" />';
         expect(parser.replace(testString)).toBe(resultString);
     });
+
+    test('No html inside the src attribute', () => {
+        const testString = '![`code`](https://example.com/image.png)';
+        const resultString = '<img src="https://example.com/image.png" alt="<code>code</code>" />';
+        expect(parser.replace(testString)).toBe(resultString);
+    })
 });
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1779,14 +1779,14 @@ describe('when should keep raw input flag is enabled', () => {
         });
     });
 });
-    
+
 test('Test code fence within inline code', () => {
     let testString = 'Hello world `(```test```)` Hello world';
     expect(parser.replace(testString)).toBe('Hello world &#x60;(<pre>test</pre>)&#x60; Hello world');
-    
+
     testString = 'Hello world `(```test\ntest```)` Hello world';
     expect(parser.replace(testString)).toBe('Hello world &#x60;(<pre>test<br />test</pre>)&#x60; Hello world');
-    
+
     testString = 'Hello world ```(`test`)``` Hello world';
     expect(parser.replace(testString)).toBe('Hello world <pre>(&#x60;test&#x60;)</pre> Hello world');
 
@@ -1893,12 +1893,9 @@ describe('Image markdown conversion to html tag', () => {
         expect(parser.replace(testString)).toBe(resultString);
     });
 
-    // Currently any markdown used inside the square brackets is converted to html string in the alt attribute
-    // The attributes should only contain plain text, but it doesn't seem possible to convert markdown to plain text
-    // or let the parser know not to convert markdown to html for html attributes
-    xtest('Image with alt text containing markdown', () => {
-        const testString = '![*bold* _italic_ ~strike~](https://example.com/image.png)';
-        const resultString = '<img src="https://example.com/image.png" alt="*bold* _italic_ ~strike~" />';
+    test('Image with alt text containing markdown', () => {
+        const testString = '![# fake-heading *bold* _italic_ ~strike~ [:-)]](https://example.com/image.png)';
+        const resultString = '<img src="https://example.com/image.png" alt="# fake-heading &ast;bold&ast; &lowbar;italic&lowbar; &#126;strike&#126; &lbrack;:-)&rbrack;" />';
         expect(parser.replace(testString)).toBe(resultString);
     });
 
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -113,12 +113,13 @@ export default class ExpensiMark {
              * Converts markdown style images to img tags e.g. ![Expensify](https://www.expensify.com/attachment.png)
              * We need to convert before linking rules since they will not try to create a link from an existing img
              * tag.
+             * Additional sanitization is done to the alt attribute to prevent parsing it further to html by later rules.
              */
             {
                 name: 'image',
                 regex: MARKDOWN_IMAGE_REGEX,
-                replacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}" alt="${g1}" />`,
-                rawInputReplacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}" alt="${g1}" data-raw-href="${g2}" data-link-variant="labeled" />`
+                replacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}" alt="${this.escapeMarkdownEntities(g1)}" />`,
+                rawInputReplacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}" alt="${this.escapeMarkdownEntities(g1)}" data-raw-href="${g2}" data-link-variant="labeled" />`
             },
 
             /**
@@ -944,5 +945,28 @@ export default class ExpensiMark {
         const linksInOld = this.extractLinksInMarkdownComment(oldComment);
         const linksInNew = this.extractLinksInMarkdownComment(newComment);
         return linksInOld === undefined || linksInNew === undefined ? [] : _.difference(linksInOld, linksInNew);
+    }
+
+    /**
+     * Replace MD characters with their HTML entity equivalent
+     * @param {String} text
+     * @return {String}
+     */
+    escapeMarkdownEntities(text) {
+        // A regex pattern matching special MD characters we'd like to escape
+        const pattern = /([*_{}[\]~])/g;
+
+        // A map of MD characters to their HTML entity equivalent
+        const entities = {
+            '*': '&ast;',
+            _: '&lowbar;',
+            '{': '&lbrace;',
+            '}': '&rbrace;',
+            '[': '&lbrack;',
+            ']': '&rbrack;',
+            '~': '&#126;',
+        };
+
+        return text.replace(pattern, char => entities[char] || char);
     }
 }
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -198,7 +198,7 @@ export default class ExpensiMark {
 
                 process: (textToProcess, replacement) => {
                     const regex = new RegExp(
-                        `(?![^<]*>|[^<>]*<\\/(?!h1>))([_*~]*?)${MARKDOWN_URL_REGEX}\\1(?!((?:(?!<a).)+)?<\/a>|[^<]*(<\/pre>|<\/code>|.+\/>))`,
+                        `(?![^<]*>|[^<>]*<\\/(?!h1>))([_*~]*?)${MARKDOWN_URL_REGEX}\\1(?!((?:(?!<a).)+)?<\\/a>|[^<]*(<\\/pre>|<\\/code>|.+\\/>))`,
                         'gi',
                     );
                     return this.modifyTextForUrlLinks(regex, textToProcess, replacement);
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -198,7 +198,7 @@ export default class ExpensiMark {
 
                 process: (textToProcess, replacement) => {
                     const regex = new RegExp(
-                        `(?![^<]*>|[^<>]*<\\/(?!h1>))([_*~]*?)${MARKDOWN_URL_REGEX}\\1(?!((?:(?!<a).)+)?<\\/a>|[^<]*(<\\/pre>|<\\/code>))`,
+                        `(?![^<]*>|[^<>]*<\\/(?!h1>))([_*~]*?)${MARKDOWN_URL_REGEX}\\1(?!((?:(?!<a).)+)?<\/a>|[^<]*(<\/pre>|<\/code>|.+\/>))`,
                         'gi',
                     );
                     return this.modifyTextForUrlLinks(regex, textToProcess, replacement);
--- a/__tests__/ExpensiMark-Markdown-test.js
+++ b/__tests__/ExpensiMark-Markdown-test.js
@@ -769,4 +769,10 @@ describe('Image tag conversion to markdown', () => {
         const resultString = '![https://example.com/image.png](https://example.com/image.png)';
         expect(parser.htmlToMarkdown(testString)).toBe(resultString);
    });
+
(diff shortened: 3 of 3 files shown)


MonilBhavsar:
Puneet, adding you as reviewer as you're assigned to the issue

MonilBhavsar:
Looks like this commit is not verified https://github.com/Expensify/expensify-common/pull/661/commits/caa6cf293bb8dd045943c413d5eeeb499ca49233

situchan:
This should be on hold for https://github.com/Expensify/expensify-common/pull/658, right?

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -1886,3 +1886,40 @@ describe('multi-level blockquote', () => {
...
+
+    test('Text containing images', () => {
+        const testString = 'An image of a banana: ![banana](https://example.com/banana.png) an image of a developer: ![dev](https://example.com/developer.png)';
+        const resultString = 'An image of a banana: <img src="https://example.com/banana.png" alt="banana" /> an image of a developer: <img src="https://example.com/developer.png" alt="dev" />';
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
+    test('Image with alt text containing markdown', () => {
+        const testString = '![# fake-heading *bold* _italic_ ~strike~ [:-)]](https://example.com/image.png)';
+        const resultString = '<img src="https://example.com/image.png" alt="&num; fake-heading &ast;bold&ast; &lowbar;italic&lowbar; &#126;strike&#126; &lbrack;:-)&rbrack;" />';
```
marcaaron, line 1905: hmm what about if the alt is:

`[cool" onerror="alert('xss')"]`

or something lol - I am trying hard to make the xss happen 😂

lib/ExpensiMark.js
```diff
@@ -114,6 +116,19 @@ export default class ExpensiMark {
...
+             * Converts markdown style images to img tags e.g. ![Expensify](https://www.expensify.com/attachment.png)
+             * We need to convert before linking rules since they will not try to create a link from an existing img
+             * tag.
+             * Additional sanitization is done to the alt attribute to prevent parsing it further to html by later rules.
+             */
+            {
+                name: 'image',
+                regex: MARKDOWN_IMAGE_REGEX,
+                replacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}" alt="${this.escapeMarkdownEntities(g1)}" />`,
+                rawInputReplacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}" alt="${this.escapeMarkdownEntities(g1)}" data-raw-href="${g2}" data-link-variant="labeled" />`
```
marcaaron, line 129: Escaping is good - but why display them at all? The alt tag will show a bunch of junk if the image fails to load or they hover over it.

marcaaron (approved):
Left a comment - not really a blocker but seems worth fixing now since we're in here.

lib/ExpensiMark.js
kidroca: I considered keeping the text plain initially, but realized that it wouldn't allow us to revert to the original Markdown message from the HTML. This way, the original format is retained.

Regarding hover text, the `alt` attribute doesn't display text on hover. For that, Markdown has an extended syntax (`alt`) that adds a `title` attribute to the resulting HTML. However, we don't support this feature, similarly to our approach with links.

If an image fails to load, the fallback is something like:
image

But NewDot and react-native-web don't display this fallback since they render a hidden `` tag instead.

__tests__/ExpensiMark-HTML-test.js
kidroca: Thanks for the idea. The general HTML escaping logic, which is enabled by default, appears to cover this case. It results in something like the following:

```html
<img src=\"https://example.com/image.png\" alt=\"test&quot; onerror=&quot;alert(&#x27;xss&#x27;)&quot;\" />
```

I've added a test just in case

marcaaron: Nice thanks!

lib/ExpensiMark.js
marcaaron: Ah hmm. Quite weird. Perhaps we can filter it before rendering i.e. not actually modify the underlying `html` field. But we can look into that in the App PR. Thanks for the explanation.

kidroca:
> Looks like this commit is not verified [caa6cf2](https://github.com/Expensify/expensify-common/commit/caa6cf293bb8dd045943c413d5eeeb499ca49233)

Fixed

> This should be on hold for #658, right?

Yes, and #658 was just merged

---

I found a bug I've described here https://github.com/Expensify/expensify-common/pull/661#discussion_r1530208382 and I'm currently trying to fix it

marcaaron:
Thanks @kidroca! 

Let us know when it's ready for review and I can give this a merge. 🙇

lib/ExpensiMark.js
```diff
@@ -926,4 +954,28 @@ export default class ExpensiMark {
...
+     * @param {String} text
+     * @return {String}
+     */
+    escapeMarkdownEntities(text) {
+        // A regex pattern matching special MD characters we'd like to escape
+        const pattern = /([*_{}[\]#~])/g;
+
+        // A map of MD characters to their HTML entity equivalent
+        const entities = {
+            '*': '&ast;',
```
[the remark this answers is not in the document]
kidroca: The aforementioned bug has been resolved by modifying the `autolink` pattern to recognize self-closing tags such as `img`. The absence of backticks is due to the fact that when `escapeMarkdownEntities` is executed, all backticks have already been transformed into `<code>` or `<pre>`

kidroca:
I've addressed the bug related to the `autolink` rule by adjusting the pattern to exclude self-closing tags, like ``. The pull request is ready for review. However, I'm cautious about declaring this the definitive fix for HTML within attributes. Given the current order of rule application, where code blocks and inline code precede others, HTML tags such as `<code> ... </code>` could still appear within the alt attribute.

This PR aimed to find a minimal solution to prevent these occurrences, but it couldn't fully achieve that, so we might consider looking for a different approach.
- we could merge the current PR as it provides some coverage
- we can turn to a broader audience for ideas
- we can test a bit to make certain there are no undetected problems

marcaaron:
>we could merge the current PR as it provides some coverage

I'd lean towards this unless there are any glaring major regressions or limitations.

kidroca:
> > Given the current order of rule application, where code blocks and inline code precede others, HTML tags such as ` ... ` could still appear within the alt attribute.
> 
> That sounds kind of edge case to me. Can you give a practical example of what a user might do and what weird behavior would result?

The latest test has an example:

```js
    test('No html inside the src attribute', () => {
        const testString = '![`code`](https://example.com/image.png)';
        const resultString = '<img src="https://example.com/image.png" alt="<code>code</code>" />';
    })
```

As you can see this particular input would result in html inside the alt attribute
The part ``[`code`]`` is first converted to html `[<code>code</code>]` and later added to the alt attribute
It doesn't seem to lead to anything breaking, but preferably should be escaped as I'm not sure it's valid 
We've we able to escape `<strong>` and alike because their rules run after the `image`...

marcaaron:
Ah I see. Well, I think it's valid HTML - but definitely leaves us with semantically strange `alt` tags. Tbh, I don't think we need to worry about this right now. We can try to address the alt tag inconsistencies as a follow up.

marcaaron (approved):
I'm gonna merge this one because I think we've already got some funky alt tags with the escape characters and creating an inline code block link doesn't seem all that common. And it's only the `alt` tag that gets affected - which seems like a separate problem from allowing inline images.

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -1934,4 +1931,10 @@ describe('Image markdown conversion to html tag', () => {
         const resultString = '<img src=\"https://example.com/image.png\" alt=\"test&quot; onerror=&quot;alert(&#x27;xss&#x27;)\" />';
         expect(parser.replace(testString)).toBe(resultString);
     });
+
+    test('No html inside the src attribute', () => {
+        const testString = '![`code`](https://example.com/image.png)';
+        const resultString = '<img src="https://example.com/image.png" alt="<code>code</code>" />';
+        expect(parser.replace(testString)).toBe(resultString);
+    })
```
tomekzaw, line 1939: @kidroca Shouldn't it be `alt="&lt;code&gt;code&lt;/code&gt;"` instead of `alt="<code>code</code>"`?

tomekzaw: > How does this impact Live Markdown functionality?

Currently, Live Markdown uses pretty basic algorithm for parsing HTML that alternates between finding `<` and then `>`.

The output `<img src="..." alt="<code>...</code>" ... />` breaks the parser as the `>` from `<code>` gets matched instead of the (self-)closing `>` from `img` tag.

According to the [HTML standard](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2), `>` is not a valid character in HTML attribute:

> Attributes have a name and a value. Attribute names must consist of one or more characters other than [controls](https://infra.spec.whatwg.org/#control), U+0020 SPACE, U+0022 ("), U+0027 ('), U+003E (>), U+002F (/), U+003D (=), and [noncharacters](https://infra.spec.whatwg.org/#noncharacter).

Is there a chance we could somehow escape the `alt` attribute?

kidroca: Hi, 
The text you're referring to in your quote is regarding the attribute **name** - an attribute name cannot contain the specified characters

> According to the [HTML standard](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2), `>` is not a valid character in HTML attribute:
> 
> > Attributes have a name and a value. Attribute names must consist of one or more characters other than [controls](https://infra.spec.whatwg.org/#control), U+0020 SPACE, U+0022 ("), U+0027 ('), U+003E (>), U+002F (/), U+003D (=), and [noncharacters](https://infra.spec.whatwg.org/#noncharacter).

Attribute values on the other hand, are less restricted as long as the value is wrapped in matching quotes. There's a [specific text in the shared spec](https://html.spec.whatwg.org/multipage/introduction.html#restrictions-on-content-models-and-on-attribute-values:~:text=The%20attribute%20value%20can%20remain%20unquoted%20if%20it%20doesn%27t%20contain%20ASCII%20whitespace%20or%20any%20of%20%22%20%27%20%60%20%3D%20%3C%20or%20%3E.%20Otherwise%2C%20it%20has%20to%20be%20quoted%20using%20either%20single%20or%20double%20quotes.) clearly stating that `<` and `>` are supported as long as the attribute value is wrapped in quotes.

> The attribute value can remain [unquoted](https://html.spec.whatwg.org/multipage/syntax.html#unquoted) if it doesn't contain [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace) or any of " ' ` = < or >. Otherwise, it has to be quoted using either single or double quotes.

In any case I'll look into it a bit further to see if there's anything I can do to mitigate the issue

tomekzaw: > The text you're referring to in your quote is regarding the attribute name 

@kidroca oops, you're right, sorry for mistake!

> In any case I'll look into it a bit further to see if there's anything I can do to mitigate the issue

I believe we could use a more complex HTML parser in Live Markdown.

cc @robertKozik who adapted ExpensiMark to Live Markdown.

robertKozik: I've checked a bit. Why don't we just escape the values in attributes (as we are already escaping markdown syntax)? I don't have much context on your PR but it seems to work quite well, at least we are not getting an error, so we would have more time to work on better solution if needed.

kidroca: > Why don't we just escape the values in attributes (as we are already escaping markdown syntax)?

The parsing process applies rules sequentially. By the time the `image` rule executes, the original markdown has been transformed into HTML and incorporated into the attribute, resulting in elements like 
`<img src="..." alt="<code>...</code>" />`. Escaping attribute values initially would lead to 
`&lt;code&gt;...&lt;/code&gt;`. This approach, however, complicates reverting HTML back to its original markdown, particularly when editing comments. Instead of restoring to the original markdown format ``[`...`](...)``, users would encounter a literal `[<code>...</code>](...)` in their edits.

---

I've identified an alternative solution that involves converting the HTML attribute back into the originally escaped markdown. You can find the implementation details here: https://github.com/Expensify/expensify-common/pull/676. It seems to work and should hopefully resolve any issues with the live markdown preview feature

robertKozik: But does reverting back when editing work? When I've tested it yesterday I think typing ` ```test``` ` as img name, and trying to edit it leads to `test` string. I may be wrong tho, as I'm not remember exactly

Yeah your alternative solution will indeed solve our issues as we wouldn't have an `<` or `>` inside attributes. What is ETA for merging it?

kidroca: We seem to be in a bit of a dead lock situation here. To advance my `expensify-common` PR, I need to demonstrate its effectiveness within the `main` branch of NewDot. Would the solution I proposed here be acceptable to you: https://github.com/Expensify/App/issues/38952#issuecomment-2035196593?

Feel free to utilize this snapshot of `expensify-common` in your projects:
```json
"expensify-common": "git+ssh://git@github.com/kidroca/expensify-common.git#3ad1c3322f4ec8999483955afe6f55c95e78172a",
```
This snapshot reflects the changes in my PR for `expensify-common`.

1. Incorporating it into your PR would resolve your issue
2. It would allow QA to test the MD Image feature. 
  - But we should note to test this in the Tests section of your PR.
3. This in turn would allow approve the merge of the `expensify-common` PR
4. Letting you to finally switch back `expensify-common` to the Expensify's repo 

I'll be ready to cover any follow-up work related to the MD image feature.

What are your thoughts?

robertKozik: It's a bit tricky because we're incorporating the `expensify-common` version inside `react-native-live-markdown` as well. So we would need to point to your branch within the new library version, which isn't my preferred approach. If possible, merging your `expensify-common` PR as it is would be great. If not, we can go with a patch containing your changes.

Here's how I envision the process:
1. Merge your `expensify-common` PR (if possible).
2. Publish a new version of `react-native-live-markdown` with both our and your fixes.
3. Create a joint PR for `react-native-live-markdown` and `expensify-common` in the E/App repo.

robertKozik: @kidroca Apart from the one problem [I've mentioned here](https://github.com/Expensify/expensify-common/pull/676#issuecomment-2037523495) everything looks great. I think we can proceed with the patch inside the `react-native-live-markdown` so it will unblock you - does it sound good for you?

PR #676, opened 2024-04-02
🔧 ExpensiMark: Markdown image shorthand syntax

kidroca:
This PR introduces refinements in the `ExpensiMark` library to enhance the handling of markdown syntax for images, including the introduction of shorthand syntax for images without alt text, and refines the cleanup of HTML attributes

### Fixed Issues
$ https://github.com/Expensify/App/issues/38952

# Tests

1. Unit tests have been updated in `__tests__/ExpensiMark-HTML-test.js` and `__tests__/ExpensiMark-Markdown-test.js` to cover the new markdown shorthand for images without alt text, the handling of images with empty and special characters in alt text, and the refinement in HTML attribute cleanup. These tests validate the parsing of markdown to HTML and vice versa, ensuring that the changes behave as expected.
1. Manual tests were conducted to verify that the new markdown shorthand for images is correctly converted to HTML, that images with and without alt text are handled properly, and that no XSS vulnerabilities are introduced through the alt text or any other HTML attributes. The tests also confirmed that existing functionality for other markdown elements remains unaffected.

# QA
1. QA needs to validate that images using the new shorthand markdown syntax are correctly rendered in the app. This includes testing with various image URLs, with and without alt text, and ensuring that images embedded within text blocks are displayed correctly.
1. It's crucial to test for regressions in the rendering of markdown content, particularly other markdown elements like links, bold/italic text, and code blocks. Ensuring that the sanitization of HTML attributes does not inadvertently affect the rendering of other elements or introduce any rendering issues is also important.

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1978,4 +1978,10 @@ describe('Image markdown conversion to html tag', () => {
         const resultString = '<img src="https://example.com/image.png" alt="&#x60;code&#x60;" />';
         expect(parser.replace(testString)).toBe(resultString);
     });
+
+    test('No html inside the alt attribute - pre tag', () => {
+        const testString = '![```code```](https://example.com/image.png)';
+        const resultString = '<img src="https://example.com/image.png" alt="&#x60;&#x60;&#x60;code&#x60;&#x60;&#x60;" data-raw-href="https://example.com/image.png" data-link-variant="labeled" />';
+        expect(parser.replace(testString, {shouldKeepRawInput: true})).toBe(resultString);
+    });
 });
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1910,15 +1910,27 @@ describe('multi-level blockquote', () => {
 });
 
 describe('Image markdown conversion to html tag', () => {
-    test('Single image', () => {
+    test('Single image with alt text', () => {
         const testString = '![test](https://example.com/image.png)';
         const resultString = '<img src="https://example.com/image.png" alt="test" />';
         expect(parser.replace(testString)).toBe(resultString);
     });
 
+    test('Single image with empty alt text', () => {
+        const testString = '![](https://example.com/image.png)';
+        const resultString = '<img src="https://example.com/image.png" />';
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
+    test('Single image short syntax without alt text', () => {
+        const testString = '!(https://example.com/image.png)';
+        const resultString = '<img src="https://example.com/image.png" />';
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
     test('Text containing images', () => {
-        const testString = 'An image of a banana: ![banana](https://example.com/banana.png) an image of a developer: ![dev](https://example.com/developer.png)';
-        const resultString = 'An image of a banana: <img src="https://example.com/banana.png" alt="banana" /> an image of a developer: <img src="https://example.com/developer.png" alt="dev" />';
+        const testString = 'An image of a banana: ![banana](https://example.com/banana.png) an image without alt: !(https://example.com/developer.png)';
+        const resultString = 'An image of a banana: <img src="https://example.com/banana.png" alt="banana" /> an image without alt: <img src="https://example.com/developer.png" />';
         expect(parser.replace(testString)).toBe(resultString);
     });
 
@@ -1965,5 +1977,5 @@ describe('Image markdown conversion to html tag', () => {
         const testString = '![`code`](https://example.com/image.png)';
         const resultString = '<img src="https://example.com/image.png" alt="<code>code</code>" />';
         expect(parser.replace(testString)).toBe(resultString);
-    })
+    });
 });
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1936,7 +1936,7 @@ describe('Image markdown conversion to html tag', () => {
 
     test('Image with alt text containing markdown', () => {
         const testString = '![# fake-heading *bold* _italic_ ~strike~ [:-)]](https://example.com/image.png)';
-        const resultString = '<img src="https://example.com/image.png" alt="# fake-heading &ast;bold&ast; &lowbar;italic&lowbar; &#126;strike&#126; &lbrack;:-)&rbrack;" />';
+        const resultString = '<img src="https://example.com/image.png" alt="# fake-heading *bold* _italic_ ~strike~ [:-)]" />';
         expect(parser.replace(testString)).toBe(resultString);
     });
 
@@ -1975,7 +1975,7 @@ describe('Image markdown conversion to html tag', () => {
 
     test('No html inside the src attribute', () => {
         const testString = '![`code`](https://example.com/image.png)';
-        const resultString = '<img src="https://example.com/image.png" alt="<code>code</code>" />';
+        const resultString = '<img src="https://example.com/image.png" alt="&#x60;code&#x60;" />';
         expect(parser.replace(testString)).toBe(resultString);
     });
 });
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -21,6 +21,13 @@ test('Test multi-line bold markdown replacement', () => {
     expect(parser.replace(testString)).toBe(replacedString);
 });
 
+
+test('Test bold within code blocks is skipped', () => {
+    const testString = 'bold\n```*not a bold*```\nThis is *bold*';
+    const replacedString = 'bold<br /><pre>*not&#32;a&#32;bold*</pre>This is <strong>bold</strong>';
+    expect(parser.replace(testString)).toBe(replacedString);
+});
+
 test('Test heading markdown replacement', () => {
     let testString = '# Heading should not have new line after it.\n';
     expect(parser.replace(testString)).toBe('<h1>Heading should not have new line after it.</h1>');
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1952,6 +1952,12 @@ describe('Image markdown conversion to html tag', () => {
         expect(parser.replace(testString, {shouldKeepRawInput: true})).toBe(resultString);
     });
 
+    test('Single short syntax image with raw data attributes', () => {
+        const testString = '!(https://example.com/image.png)';
+        const resultString = '<img src="https://example.com/image.png" data-raw-href="https://example.com/image.png" data-link-variant="auto" />';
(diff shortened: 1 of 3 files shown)


lib/ExpensiMark.js
```diff
@@ -959,27 +983,4 @@ export default class ExpensiMark {
...
-            _: '&lowbar;',
-            '{': '&lbrace;',
-            '}': '&rbrace;',
-            '[': '&lbrack;',
-            ']': '&rbrack;',
-            '~': '&#126;',
-        };
-
-        return text.replace(pattern, char => entities[char] || char);
-    }
```
kidroca, line 984: With the introduction of `attributeCleanup` rule - this logic is no longer needed

lib/ExpensiMark.js
```diff
@@ -3,9 +3,8 @@ import Str from './str';
 import {MARKDOWN_URL_REGEX, LOOSE_URL_REGEX, URL_REGEX} from './Url';
 import {CONST} from './CONST';
 
-const MARKDOWN_LINK = `\\[([^\\][]*(?:\\[[^\\][]*][^\\][]*)*)]\\(${MARKDOWN_URL_REGEX}\\)(?![^<]*(<\\/pre>|<\\/code>))`;
-const MARKDOWN_LINK_REGEX = new RegExp(MARKDOWN_LINK, 'gi');
-const MARKDOWN_IMAGE_REGEX = new RegExp(`\\!${MARKDOWN_LINK}`, 'gi');
+const MARKDOWN_LINK_REGEX = new RegExp(`\\[([^\\][]*(?:\\[[^\\][]*][^\\][]*)*)]\\(${MARKDOWN_URL_REGEX}\\)(?![^<]*(<\\/pre>|<\\/code>))`, 'gi');
+const MARKDOWN_IMAGE_REGEX = new RegExp(`\\!(?:\\[([^\\][]*(?:\\[[^\\][]*][^\\][]*)*)])?\\(${MARKDOWN_URL_REGEX}\\)(?![^<]*(<\\/pre>|<\\/code>))`, 'gi');
```
kidroca, line 7: MARKDOWN_LINK's string was directly moved into the MARDOWN_LINK_REGEX constructor as it's no longer reusable 

`MARKDOWN_IMAGE_REGEX` was updated and although is similar to `MARKDOWN_LINK` the `[alt text]` part was made optional

lib/ExpensiMark.js
```diff
@@ -330,6 +330,24 @@ export default class ExpensiMark {
...
+                // We're limiting the scope of this rule to 'img' tags and their 'alt' attributes, which are the
+                // primary sources of such issues. This approach keeps the operation efficient and focused.
+                // However, the pattern is extensible to additional tags and attributes as needed.
+                name: 'attributeCleanup',
+                regex: /<(?:img)[^>]*\s(?:alt)=(["'])([^"']*)\1[^>]*>/gi,
+                replacement: (match, g1, g2) => {
+                    const cleanedG2 = this.htmlToMarkdown(g2);
+                    if (cleanedG2 === g2) {
+                        return match;
+                    }
```
chiragsalian, line 346: Do we need this? What happens if we always return `match.replace(g2, _.escape(cleanedG2))` instead? 
Will it be a performance issue or something else?
Because if its a performance issue wont it be faster to check if there are any markdowns to convert if none return g2 early. But if its not a performance concern won't it be more readable skipping this and always returning `match.replace(g2, _.escape(cleanedG2))`.

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -1910,21 +1910,33 @@ describe('multi-level blockquote', () => {
...
-        const resultString = 'An image of a banana: <img src="https://example.com/banana.png" alt="banana" /> an image of a developer: <img src="https://example.com/developer.png" alt="dev" />';
+        const testString = 'An image of a banana: ![banana](https://example.com/banana.png) an image without alt: !(https://example.com/developer.png)';
+        const resultString = 'An image of a banana: <img src="https://example.com/banana.png" alt="banana" /> an image without alt: <img src="https://example.com/developer.png" />';
         expect(parser.replace(testString)).toBe(resultString);
     });
 
     test('Image with alt text containing markdown', () => {
         const testString = '![# fake-heading *bold* _italic_ ~strike~ [:-)]](https://example.com/image.png)';
-        const resultString = '<img src="https://example.com/image.png" alt="# fake-heading &ast;bold&ast; &lowbar;italic&lowbar; &#126;strike&#126; &lbrack;:-)&rbrack;" />';
+        const resultString = '<img src="https://example.com/image.png" alt="# fake-heading *bold* _italic_ ~strike~ [:-)]" />';
```
chiragsalian, line 1939: have you tested this and the next change with our backend code? I'm mostly wondering if our backend would support these changes just fine or not.

chiragsalian (dismissed):
Code LGTM, i just have 2 questions for you. If nothing needs to be modified I'll merge once answered.

__tests__/ExpensiMark-HTML-test.js
kidroca: I see where you're coming from, and here's what I can tell you:

1. I wasn't aware that this package is utilized in the backend as well, mainly because I don't have access to Expensify's private repositories.
2. Regarding your concern - we're using "allowed" characters verbatim instead of escaping them, but we still escape characters that are considered "illegal."

Does this address your concerns, or is there something specific you're worried about regarding backend compatibility?

lib/ExpensiMark.js
kidroca: Since `g2` can include HTML content, and I'm not aware of a quick function to check for HTML markup, utilizing the parser allows us to both detect HTML and process the content for the next step efficiently.
The code, while a bit verbose, optimizes performance slightly - the `_.escape()` function is called only when necessary.

If you wish to shorten it to one line, just say so and I'll update it when I get the chance

robertKozik:
Hi! @kidroca 
Can you check the case when label is a code block and shouldKeepRawInput is set to true? Seems wrong to me
```
    test.only('No html inside the alt attribute - pre tag', () => {
        const testString = '![```code```](https://example.com/image.png)';
        const resultString = '<img src="https://example.com/image.png" alt="&#x60;&#x60;&#x60;code&#x60;&#x60;&#x60;" data-raw-href="https://example.com/image.png" data-link-variant="labeled" />';
        expect(parser.replace(testString, {shouldKeepRawInput: true})).toBe(resultString);
    });
```

kidroca:
@robertKozik, @chiragsalian, I have updated the pull request with a revised approach to handling HTML attributes. The original method aimed to clean up HTML within attributes but proved unreliable because distinguishing embedded HTML from the attribute content was challenging. Consider the example:

```js
<img src="https://example.com/image.png" alt="<pre data-code-raw="code">code</pre>" data-raw-href="https://example.com/image.png" data-link-variant="labeled" /> />
```

In this case, the `alt` attribute includes additional quotes that complicate detecting the attribute's end due to the embedded HTML. To address this, I've modified the Regex patterns for the `italic`, `bold`, and `strike` rules to prevent matches within HTML attributes. Additionally, I have added tests to cover these scenarios (thanks to @robertKozik for the suggestion), and all existing tests for these rules continue to pass.

As the scope and focus of the PR have shifted, I kindly request a re-review. I have made an effort to limit the extent of changes for easier review.

kidroca:
@robertKozik, regarding:
> include Expensify/expensify-common#676 PR as patch package

While using a patch package might seem convenient, I've updated the PR, which necessitates creating the patch again. 

Initially, my suggestion to use the commit hash from my fork was intended as a temporary measure for testing purposes in your PR. This approach would enable us to validate the integration, merge the changes into expensify-common, and then revert to the original package source.
(Here's the hash if you've reconsidered: `"expensify-common": "git+ssh://git@github.com/kidroca/expensify-common.git#7a7c075bff9a07eb280fc720bec65b6d2acab161"`)

Alternatively, if it’s more feasible, I can open a new PR in NewDot using a previous version of `live-markdown`. Let me know if you prefer this method, and I can set it up later.

lib/ExpensiMark.js
```diff
@@ -118,13 +117,14 @@ export default class ExpensiMark {
...
+             * Additional sanitization is done to the alt attribute to prevent parsing it further to html by later
+             * rules.
              */
             {
                 name: 'image',
                 regex: MARKDOWN_IMAGE_REGEX,
-                replacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}" alt="${this.escapeMarkdownEntities(g1)}" />`,
-                rawInputReplacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}" alt="${this.escapeMarkdownEntities(g1)}" data-raw-href="${g2}" data-link-variant="labeled" />`
+                replacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}"${g1 ? ` alt="${this.escapeAttributeContent(g1)}"` : ''} />`,
+                rawInputReplacement: (match, g1, g2) => `<img src="${Str.sanitizeURL(g2)}"${g1 ? ` alt="${this.escapeAttributeContent(g1)}"` : ''} data-raw-href="${g2}" data-link-variant="labeled" />`
```
robertKozik, line 127: One minor thing, sorry for bother @kidroca .
If user types `!(link)` data-link-variant should be set to "auto" rather than "labeled" indicating that it's typed using shorthand syntax

This is needed to distinguish between `` and `!(link)`

robertKozik:
Actually I'm still prefering patching the repo with your changes, rather than pointing to the forked repository.
I'll merge it and open the Expensify PR which will unblock you.
I've added one more comment - do you think it's solvable? @kidroca

kidroca:
> I've added one more comment - do you think it's solvable? @kidroca

Thanks, I'll post an update in a minute

robertKozik:
I've came up with something like that:
`data-link-variant="${typeof(g1) === 'string' ? 'labeled': 'auto'}"`

kidroca:
Updated

robertKozik:
@kidroca  [This PR](https://github.com/Expensify/App/pull/39905) will unblock you

robertKozik (approved):

robertKozik:
Pls merge main, bold -> <strong/> seems to go under the radar in HTML attributes

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -21,6 +21,13 @@ test('Test multi-line bold markdown replacement', () => {
     expect(parser.replace(testString)).toBe(replacedString);
 });
 
+
+test('Test bold within code blocks is skipped', () => {
+    const testString = 'bold\n```*not a bold*```\nThis is *bold*';
+    const replacedString = 'bold<br /><pre>*not&#32;a&#32;bold*</pre>This is <strong>bold</strong>';
+    expect(parser.replace(testString)).toBe(replacedString);
+});
```
kidroca, line 29: Added this test to ensure the merge conflict was resolved correctly
The test is related to code added in: https://github.com/Expensify/expensify-common/pull/681

kidroca:
Merged `main` there was a conflict related to `bold` handling, but it's resolved

chiragsalian (approved):

PR #680, opened 2024-04-09
[Mentions v2] Add new rule in ExpensiMark for room mentions

robertKozik:
> [!IMPORTANT]
> Room mentions consists of `#` symbol followed by room name. Room name consist of any char sequence connected by hyphens (no spaces is allowed inside room name)

### Fixed Issues
$ https://github.com/Expensify/App/issues/39538

# Tests
I've created multiple unit tests to cover most of the possible usages of room/report mentions. Room mentions consists of `#` symbol followed by room name 
Check if existing unit tests are not failing after my. changes, added additional tests to cover room mentions

# QA
Room mentions are not yet visible inside E/App, so to validate this changes we should only check for regressions or other markdown rules
Room mention syntax is very similar to Heading syntax. Extensive testing of this aspeect of markdown parser is recommended. Moreover all other markdown rules should be untouched

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1959,5 +1959,92 @@ describe('Image markdown conversion to html tag', () => {
         const testString = '![`code`](https://example.com/image.png)';
         const resultString = '<img src="https://example.com/image.png" alt="<code>code</code>" />';
         expect(parser.replace(testString)).toBe(resultString);
-    })
+    });
+});
+
+describe('room mentions', () => {
+    test('simple room mention', () => {
+        const testString = '#room';
+        const resultString = '<mention-room>#room</mention-room>';
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
+    test('room mention with leading word and space', () => {
+        const testString = 'hi all #room';
+        const resultString = 'hi all <mention-room>#room</mention-room>';
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
+    test('room mention with leading word and no space', () => {
+        const testString = 'hi all#room';
+        const resultString = 'hi all#room';
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
+    test('room mention with space between hash and room name', () => {
+        const testString = '# room';
+        const resultString = '<h1>room</h1>';
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
+    test.only('room mention with markdown syntax before the # prefix', () => {
+        const testString = 'hello *#room';
+        const resultString = 'hello *<mention-room>#room</mention-room>';
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
+    test('room mention with italic, bold and strikethrough styles', () => {
+        const testString = '#room'
+        + ' _#room_'
+        + ' *#room*'
+        + ' ~#room~'
+        + ' [#room](google.com)'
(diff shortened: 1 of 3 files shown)


lib/ExpensiMark.js
```diff
@@ -169,6 +169,13 @@ export default class ExpensiMark {
                 },
             },
 
+            {
```
rlinoz, line 172: Can we add a comment here explaining the regex the same way the other mention regex have?

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -1959,5 +1959,92 @@ describe('Image markdown conversion to html tag', () => {
         const testString = '![`code`](https://example.com/image.png)';
         const resultString = '<img src="https://example.com/image.png" alt="<code>code</code>" />';
         expect(parser.replace(testString)).toBe(resultString);
-    })
+    });
+});
+
+describe('room mentions', () => {
```
rlinoz, line 1965: I think we can add a test for:
- `#` without following characters, should yield no mention
- Room name with numbers
- Room name with hyphen

rlinoz (changes requested):

robertKozik:
Thanks @rlinoz, I've added tests and comment

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -1959,5 +1959,110 @@ describe('Image markdown conversion to html tag', () => {
...
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
+    test('room mention with space between hash and room name', () => {
+        const testString = '# room';
+        const resultString = '<h1>room</h1>';
+        expect(parser.replace(testString)).toBe(resultString);
+    });
+
+    test.only('room mention with markdown syntax before the # prefix', () => {
```
Beamanator, line 1990: Did you mention to have `.only` here? I'm not entirely sure what this is used for, but it looks like we don't do that anywhere else in our tests in this repo 🤔 

```suggestion
    test('room mention with markdown syntax before the # prefix', () => {
```

Beamanator (changes requested):
Can we also add at least 1 test for room names with `-` since that will probably be a pretty common use-case? Ex: `#room-name`, maybe even `#room-123`

__tests__/ExpensiMark-HTML-test.js
robertKozik: Yeah, I've commited it by mistake. Thanks for noticing it 🙇🏼

Beamanator: Cool, just checking! Thanks for the quick fix!

Beamanator (approved):
Looking good to me!

rlinoz (approved):

lib/ExpensiMark.js
```diff
@@ -169,6 +169,19 @@ export default class ExpensiMark {
...
+            /**
+             * A room mention is a string that starts with the '#' symbol and is followed by a valid room name.
+             *
+             * Note: We are allowing mentions in a format of #room-name The room name can contain any
+             * combination of letters and hyphens
+             */
+            {
+                name: 'roomMentions',
+
+                regex: /(?<![^ \n*~_])(#[\p{Ll}0-9-]{1,80})/gmiu,
```
rojiphil, line 181: NAB but certain improvements were introduced as mentioned in [this issue](https://github.com/Expensify/App/issues/46299) with respect to identifying room mentions.

PR #686, opened 2024-04-18
Pass extra data (fe. accountIDs) to HTMLToMarkdown method & add possibility to disable certain rules

robertKozik:
This PR consists of two different changes
1.  `htmlToMarkdown` method now has additional, optional argument `extras` this argument adds the possibility to pass extra data which can be used in replacement function. For now is used to pass accountID to name and reportID to name maps
2. `replace` method has additional object key `disabledRules`. Every rule name from this array will be skipped in parsing.

### Fixed Issues
$ https://github.com/Expensify/App/issues/40480
https://github.com/Expensify/App/issues/39550

# Tests
The changes, which this PR introduces, allows to more precisely interact with parser. To be precise, now we will be able to pass application state to the parser itself. We wrote couple of unit tests which mocks passing `extras` object with account ID and reportID maps to ensure that parsing shortened room and user mentions tags works as intended.

We also tested our changes inside E/App. It's not yet supported on main branch, but all changes worked as intended

# QA
These changes cannot be tested on E/App yet. Only follow-up PR will be using these changes
As we were touching methods responsible for both HTML –> markdown and markdown -> HTML we should look for regressions in every rule and parsing.

### DIFF
--- a/lib/ExpensiMark.d.ts
+++ b/lib/ExpensiMark.d.ts
@@ -1,10 +1,11 @@
-declare type Replacement = (...args: string[]) => string;
+declare type Replacement = (...args: string[], extras?: ExtrasObject) => string;
 declare type Name =
     | 'codeFence'
     | 'inlineCodeBlock'
     | 'email'
     | 'link'
     | 'hereMentions'
+    | 'roomMentions'
     | 'userMentions'
     | 'autoEmail'
     | 'autolink'
@@ -31,6 +32,11 @@ declare type Rule = {
     replacement: Replacement | string;
     pre?: (input: string) => string;
     post?: (input: string) => string;
+};
+
+declare type ExtrasObject = {
+    reportIdToName?: Record<string, string>;
+    accountIDToName?: Record<string, string>;
 };
 export default class ExpensiMark {
     rules: Rule[];
@@ -91,14 +97,14 @@ export default class ExpensiMark {
      * @param htmlString
      * @param extras
      */
-    htmlToMarkdown(htmlString: string, extras?: Object): string;
+    htmlToMarkdown(htmlString: string, extras?: ExtrasObject): string;
     /**
      * Convert HTML to text
      *
      * @param htmlString
      * @param extras
      */
-    htmlToText(htmlString: string, extras?: Object): string;
+    htmlToText(htmlString: string, extras?: ExtrasObject): string;
     /**
      * Modify text for Quotes replacing chevrons with html elements
      *
--- a/lib/ExpensiMark.d.ts
+++ b/lib/ExpensiMark.d.ts
@@ -7,6 +7,7 @@ declare type Name =
     | 'hereMentions'
     | 'roomMentions'
     | 'userMentions'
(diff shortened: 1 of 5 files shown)


war-in:

lib/ExpensiMark.js
```diff
@@ -453,7 +457,31 @@ export default class ExpensiMark {
 
                     return `!(${g2})`;
                 }
-            }
+            },
+            {
+                name: 'roomMention',
```
war-in, line 462: Now I wonder if we should keep the `room` here or change it to the `report`

lib/ExpensiMark.js
```diff
@@ -768,10 +836,11 @@ export default class ExpensiMark {
      * Replaces HTML with markdown
      *
      * @param {String} htmlString
+     * @param {Object} extras
```
war-in, line 839: `ExtrasObject`? 🤔

lib/ExpensiMark.js
```diff
@@ -795,13 +867,16 @@ export default class ExpensiMark {
      * Convert HTML to text
      *
      * @param {String} htmlString
+     * @param {Object} extras
```
war-in, line 870: Same here

lib/ExpensiMark.js
```diff
@@ -795,13 +867,16 @@ export default class ExpensiMark {
...
+     *
      * @returns {String}
      */
-    htmlToText(htmlString) {
+    htmlToText(htmlString, extras) {
         let replacedText = htmlString;
 
         this.htmlToTextRules.forEach((rule) => {
-            replacedText = replacedText.replace(rule.regex, rule.replacement);
+            const replacementFunction = typeof rule.replacement === 'function' ? (...args) => rule.replacement(...args, extras) : rule.replacement;
```
war-in, line 878: Should we add this comment here too?

```
// if replacement is a function, we want to pass optional extras to it
```

lib/ExpensiMark.js
robertKozik: change to report :P

lib/ExpensiMark.d.ts
```diff
@@ -1,11 +1,13 @@
-declare type Replacement = (...args: string[]) => string;
+declare type Replacement = (...args: string[], extras?: ExtrasObject) => string;
 declare type Name =
     | 'codeFence'
     | 'inlineCodeBlock'
     | 'email'
     | 'link'
     | 'hereMentions'
+    | 'roomMentions'
```
rlinoz, line 8: I'm confused, why we need the `roomMentions` here if we renamed it o `reportMentions`?

rlinoz (commented):
Looks good, just one question

lib/ExpensiMark.d.ts
robertKozik: Thanks! Missed that while implementing

rlinoz (approved):
Thanks!

rlinoz:
Merging to unblock other issues

PR #691, opened 2024-05-07
Recognise only blockquotes followed by space

robertKozik:
image

This PR changes quote regex to enforce following space after `>`.
`>` not followed by space won't be parsed as quote. In the multiple level quote context spaces between `>` signs are irrelevent

### Fixed Issues
$ https://github.com/Expensify/App/issues/41696

# Tests
I've Adjusted already present unit tests 
I've run unit tests and checked the changes in both example app of `react-native-live-markdown` and E/App itself, by bumping the expensify-common version to my branch head

# QA
bump expensify-common version in E/App and use chat normally - sent messages should be parsed with the new rule
All quotes usage are vulnerable to regressions

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1103,6 +1103,35 @@ test('Test quotes markdown replacement with text includes multiple spaces', () =
     expect(parser.replace(quoteTestStartString)).toBe(quoteTestReplacedString);
 });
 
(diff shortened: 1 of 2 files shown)


thienlnam (commented):
Thanks, this looks good - could we also add tests for the negative case where the >hello doesn't turn into a blockquote?

robertKozik:
Sure, I'm on it 👀

thienlnam (approved):
Awesome, thanks!

PR #696, opened 2024-05-14
Add missing default value to `extras`

war-in:
### Fixed Issues
$ [42161](https://github.com/Expensify/App/issues/42161)

# Tests

# QA

### DIFF
--- a/lib/ExpensiMark.js
+++ b/lib/ExpensiMark.js
@@ -2,6 +2,7 @@ import * as _ from 'underscore';
 import Str from './str';
 import * as Constants from './CONST';
 import * as UrlPatterns from './Url';
+import Log from './Log';
 
 const MARKDOWN_LINK_REGEX = new RegExp(`\\[([^\\][]*(?:\\[[^\\][]*][^\\][]*)*)]\\(${UrlPatterns.MARKDOWN_URL_REGEX}\\)(?![^<]*(<\\/pre>|<\\/code>))`, 'gi');
 const MARKDOWN_IMAGE_REGEX = new RegExp(`\\!(?:\\[([^\\][]*(?:\\[[^\\][]*][^\\][]*)*)])?\\(${UrlPatterns.MARKDOWN_URL_REGEX}\\)(?![^<]*(<\\/pre>|<\\/code>))`, 'gi');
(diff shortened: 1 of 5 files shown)


lib/ExpensiMark.js
```diff
@@ -473,7 +473,9 @@ export default class ExpensiMark {
                 replacement: (match, g1, offset, string, extras) => {
                     const reportToNameMap = extras.reportIdToName;
                     if (!reportToNameMap || !reportToNameMap[g1]) {
-                        return '';
+                        // eslint-disable-next-line no-console
+                        console.warn(`[ExpensiMark] Missing report name for ID: ${  g1}`);
```
iwiznia, line 477: Oh, this is not what I meant, we don't use console like this (as the linter is telling you). You should log using the `Log` class. Also please change them all to `alert` instead of warn.

iwiznia (changes requested):

lib/ExpensiMark.js
robertKozik: In other places in `ExpensiMark.js` there is no `Log` class usages - only `console.warn` is used twice. That's why we have done it this way. I can change it though if you think that `Log` is more suitable in this case

war-in: Actually, we use. In expensimark there are some console.warn s. That's why I've used it. Can change if you want :)

iwiznia: hmmmm not sure why we do that, but in this case, we want the log to go to our servers or we will never know it happened, since we don't have access to user's consoles.

robertKozik: I've checked quickly and I don't see any usage of `Log` from `expensify-common` in E/App as well . Can we for now stay with the `console.warn` and check it as followup so we will fix the issue with missing extras quicker?

iwiznia: We might not be using it yet, but it exists, no? Is it causing any problems if you use it? Not sure I understand the pushback...

robertKozik: I just don't want to introduce any regression as `Log` is not yet used in `ExpensiMark`.
It's not a problem to change it, let me check more extensively in E/App. I'll ping you once I'll do it.

robertKozik: Seems to work fine, but tests are failing - I'm on it right now

lib/ExpensiMark.js
```diff
@@ -473,7 +474,8 @@ export default class ExpensiMark {
                 replacement: (match, g1, offset, string, extras) => {
                     const reportToNameMap = extras.reportIdToName;
                     if (!reportToNameMap || !reportToNameMap[g1]) {
-                        return '';
+                        Log.alert(`[ExpensiMark] Missing report name for ID: ${g1}`);
```
iwiznia, line 477: Variable parameters should go in the 2nd parameter of the call, so:
```suggestion
                        Log.alert('[ExpensiMark] Missing report name', {reportID: g1});
```

iwiznia (changes requested):

lib/ExpensiMark.js
robertKozik: resolved 👌🏼

iwiznia (approved):

PR #699, opened 2024-05-17
[TS migration] Setup typescript in expensify-common & Migrate CONST, fastMerge, Device, Logger to Typescript

blazejkustra:
Going forward expensify-common code can be written in Typescript, we have to take into consideration transpiling step now.

TS code will live in `lib` catalog and transpiled code will end up in `dist` catalog. The `lib` catalog won’t be published to npm; instead, we’ll only publish the `dist` catalog.

⚠️ This means that from now on `expensify-common` imports will look like this:
```ts
// Before:
import ExpensiMark from 'expensify-common/lib/ExpensiMark';

// After:
import {ExpensiMark, Str} from 'expensify-common';
// or
import ExpensiMark from 'expensify-common/dist/ExpensiMark';
```

[Here](https://github.com/Expensify/react-native-onyx/pull/452) is a similar PR that prepared Onyx for a similar migration.

### Fixed Issues
$ https://github.com/Expensify/App/issues/42245


lib/ReportHistoryStore.jsx
```diff
@@ -263,9 +263,6 @@ export default class ReportHistoryStore {
             delete this.cache[reportID];
         }
 
-        // We'll poll the API for the un-cached history
-        const cachedHistory = this.cache[reportID] || [];
-
```
jnowakow, line 268: This change is connected to TS migration?

jnowakow (approved):
LGTM

lib/ReportHistoryStore.jsx
blazejkustra: It was unused, so there was an ESLint error

kubabutkiewicz (approved):

lib/Logger.ts
```diff
@@ -74,29 +90,29 @@ export default class Logger {
...
+     * @param includeStackTrace Must be disabled for testing
      */
-    alert(message, parameters = {}, includeStackTrace = true) {
+    alert(message: string, parameters: Parameters = {}, includeStackTrace = true) {
         const msg = `[alrt] ${message}`;
         const params = parameters;
 
         if (includeStackTrace) {
-            params.stack = JSON.stringify(new Error().stack);
+            (params as Record<string, unknown>).stack = JSON.stringify(new Error().stack);
```
fabioh8010, line 115: Shouldn't we check if `params` is a object instead of doing the assertion? It would be safer

lib/Templates.jsx
```diff
@@ -119,22 +121,18 @@ export default (function () {
          * @return {String}
          */
         get(templatePath, data = {}) {
```
fabioh8010, line 123: Why was the `try catch` removed?

lib/Logger.ts
blazejkustra: Adjusted 👍

lib/Templates.jsx
blazejkustra: Eslint rule was complaining, try catch that always throws is unnecessary.

carlosmiceli:
@tgolen I thought I'd bring you in to review this one, since I may not be the best suited to confirm that everything is as it should... I haven't seen anything wrong in the code, but I think a second more veteran opinion is worth it here.

fabioh8010 (approved):

tgolen:
I'm really sorry. I was not able to review this today due to being heads-down getting comp review ready. I will be OOO for the next 1.5 week, so I will assign someone else to review.

.eslintrc.js
```diff
@@ -1,11 +1,72 @@
...
+                        ts: 'never',
+                        tsx: 'never',
+                    },
+                ],
+            },
+        },
+        {
+            files: ['*.ts', '*.tsx'],
+            extends: ['expensify', 'plugin:@typescript-eslint/recommended', 'plugin:@typescript-eslint/stylistic', 'plugin:import/typescript', 'prettier', 'plugin:prettier/recommended'],
+            plugins: ['react', 'import', '@typescript-eslint'],
```
roryabraham, line 36: we should also add `plugin:you-dont-need-lodash-underscore/all`. I know it makes the diff much bigger but we need to drop lodash/underscore from ExpensiMark for react-native-live-markdown, so we might as well stick with our standard TS style here

lib/Logger.ts
```diff
@@ -1,35 +1,51 @@
...
+
+    constructor({serverLoggingCallback, isDebug, clientLoggingCallback}: LoggerOptions) {
         // An array of log lines that limits itself to a certain number of entries (deleting the oldest)
         this.logLines = [];
         this.serverLoggingCallback = serverLoggingCallback;
         this.clientLoggingCallback = clientLoggingCallback;
         this.isDebug = isDebug;
 
         // Public Methods
+        // eslint-disable-next-line no-constructor-return
```
roryabraham, line 27: Curious why we'd need to disable this lint rule? Couldn't we just change this to:

```ts
this.info = this.info.bind(this);
this.alert = this.alert.bind(this);
this.warn = this.warn.bind(this);
this.hmmm = this.hmmm.bind(this);
this.client = this.client.bind(this);
```

?

lib/ExpenseRule.jsx
```diff
@@ -21,11 +21,7 @@ export default class ExpenseRule {
      * @return {Object}
      */
     getApplyWhenByField(field) {
-        return (
-            _.find(this.applyWhen, (conditions) => {
-                return conditions.field === field;
-            }) || {}
-        );
+        return _.find(this.applyWhen, (conditions) => conditions.field === field) || {};
```
roryabraham, line 24: Why underscore?

roryabraham (changes requested):
Also, what is the plan for publishing this? Right now, expensify-common is consumed via a git hash in E/App's package.json

That could work if we commit `dist`, but I see you've addd it to the `.gitignore`. So what's the plan? If we need to publish this to npm can you copy the workflow from react-native-onyx?

.eslintrc.js
blazejkustra: Sure! I'm going to add it today 😄

lib/Logger.ts
blazejkustra: I didn't want to cause regressions with a change like this, but this should be easily testable, gonna do it today.

lib/ExpenseRule.jsx
blazejkustra: Because it's a JS file, and underscore and lodash is removed only from TS files

blazejkustra: It's a JS file, and we only remove `underscore`/`lodash` from TS files, right?

blazejkustra:
> Also, what is the plan for publishing this? Right now, expensify-common is consumed via a git hash in E/App's package.json
>
> That could work if we commit dist, but I see you've addd it to the .gitignore. So what's the plan? If we need to publish this to npm can you copy the workflow from react-native-onyx?

I think copying the workflow from `react-native-onyx` makes the most sense, so that we consume the package from npm.

blazejkustra:
@roryabraham I copied over the `publish` workflow from Onyx, I think the last thing we need is to add missing secrets. Would you be able to copy it over from Onyx?
- `OS_BOTIFY_COMMIT_TOKEN`
- `LARGE_SECRET_PASSPHRASE`
- `NPM_TOKEN`

FYI: [There is `expensify-common` in npm's registry already](https://www.npmjs.com/package/expensify-common).

blazejkustra:
Also, please have another look at the code, I changed it slightly! 🚀

.eslintrc.js
```diff
@@ -44,7 +52,7 @@ module.exports = {
                 'react/require-default-props': 'off',
                 'valid-jsdoc': 'off',
                 'es/no-optional-chaining': 'off',
-                'es/no-nullish-coalescing': 'off',
+                'es/no-nullish-coalescing-operators': 'off',
```
fabioh8010, line 55: What's the difference between the two ones?

fabioh8010 (approved):

.eslintrc.js
blazejkustra: `es/no-nullish-coalescing` doesn't exist, I added it by mistake.

lib/ExpenseRule.jsx
roryabraham: oh, sorry I missed that 🤦🏼 you're right

roryabraham (approved):
LGTM but let's get those secrets added before we merge

roryabraham:
Created [internal ticket](https://github.com/Expensify/Expensify/issues/399706) to add those GitHub secrets

roryabraham:
Let's see if publishing works as expected... https://github.com/Expensify/expensify-common/actions/runs/9261388961

roryabraham:
Looks like that worked (after a few more minor adjustments) 🎉

blazejkustra:
Awesome! Thanks @roryabraham for the help on this one 🙌

PR #690, opened 2024-04-29
Change accepted syntax for code fence

BartoszGrajdek:
We're currently working on adding styles to `code fence` & `inline code` blocks to Live Markdown. After some [discussion](https://swmansion.slack.com/archives/C06BDSWLDPB/p1713274778358479) regarding `code fence`/`pre` block we agreed that the best way to handle it is to limit the accepted syntax variations.

This PR adds these changes directly to ExpensiMark 🙌🏻 

### Fixed Issues
$ [GH_LINK](https://github.com/Expensify/App/issues/39518)

# Tests
1. What unit/integration tests cover your change? What autoQA tests cover your change?
All of the tests that have changed in this PR
1. What tests did you perform that validates your changed worked?
I tested it both in Live Markdown & in the Expensify App to make sure that it didn't break anything

# QA
1. What does QA need to do to validate your changes?
Check if there's just one syntax valid for `pre` blocks which should be:
```
(triple backticks)
Content...
(triple backticks)
```
3. What areas to they need to test for regressions?
Mainly Expensify chat & how it treats messages sent with different syntax variations. Live Markdown will include these changes after we get this PR merged 👀 

https://github.com/Expensify/expensify-common/assets/41169160/2f944c36-ae97-4b81-9fe6-b809367fd8cc

PR #705, opened 2024-05-24
Remove data-code-raw for `pre` blocks

BartoszGrajdek:
After we [changed the syntax for codeFence](https://github.com/Expensify/expensify-common/pull/690) we found a bug where it was possible to add a room mention inside of it. This PR prevents that from happening & removes `data-code-raw` which is not necessary after all of our changes.

### Fixed Issues
[$ GH_LINK](https://github.com/Expensify/App/issues/39518)

# Tests
All of the tests that were edited
I checked if there were any new errors inside of `react-native-live-markdown` library and inside of E/App (which was not necessary since these changes don't affect it).

# QA
Don't know actually, this is a clean-up rather than a feature
I've added some regression test to check for room mentions inside of codeFence

image


BartoszGrajdek:
cc @dangrous since he was responsible for reviewing [the previous PR](https://github.com/Expensify/expensify-common/pull/690) and has context on what's happening here 👀

youssef-lr (approved):
LGTM

dangrous (commented):
Much cleaner! Do we have a test in there that checks that urls inside <pre>s work properly? Other than that I think this is good to go.

PR #709, opened 2024-05-28
Fix/global objects

blazejkustra:
This is a follow up PR to the [setup of TS in this repo](https://github.com/Expensify/expensify-common/pull/699). While testing in the E/App [here](https://github.com/Expensify/App/pull/42387) I noticed two issues:
1. `Url` export is wrong
2. Importing from a main file `index.ts` causes other exported files to run (for example `BrowserDetect.jsx`). This causes problems because code in `expensify-common` assumes that `window` and `navigator` global objects exist, which isn't always true.

### Fixed Issues
$ https://github.com/Expensify/App/issues/42245

### DIFF
--- a/lib/API.jsx
+++ b/lib/API.jsx
@@ -8,6 +8,7 @@ import _ from 'underscore';
 // Use this deferred lib so we don't have a dependency on jQuery (so we can use this module in mobile)
 import {Deferred} from 'simply-deferred';
 import ExpensifyAPIDeferred from './APIDeferred';
+import {isWindowAvailable} from './utils';
 
 /**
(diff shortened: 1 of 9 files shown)


lib/ExpensiMark.js
```diff
@@ -483,10 +483,10 @@ export default class ExpensiMark {
                             Log.alert('[ExpensiMark] Missing account name', {accountID: g1});
                             return '@Hidden';
                         }
-
```
blazejkustra, line 486: There was a prettier diff on main

Skalakid (approved):
LGTM 💥

PR #718, opened 2024-06-12
Remove underscore from expensify-common and reduce usage of lodash and jQuery

Skalakid:
Rebased version of https://github.com/Expensify/expensify-common/pull/704 PR, that were made to sign all commits. [Slack thread](https://swmansion.slack.com/archives/C06BDSWLDPB/p1718208445619269?thread_ts=1716410498.094849&cid=C06BDSWLDPB)

The original goal of this PR was to remove `underscore`, `jQuery`, and `lodash` from files that are connected to `ExpensiMark` so we can workletize these functions and enable the `react-native-live-markdown` to run parser on UI thread. We wanted to block usage of these libraries in the whole repo, however without knowing the use cases and without being able to test it inside the private Expensify reposthat uses some of these files, we decided to:
- remove whole `underscore` from repo, since it's quite easy to do without testing
- replace more complex `underscore` fucntions with `lodash` versions because it doesn;t make. sene to re-write them
- remove all these 3 libraries from: `ExpensiMark`, `Logger`, `Src` so we can workletize functions from these files and run them on UI thread in  `react-native-live-markdown`

### Fixed Issues
https://github.com/Expensify/App/issues/42494

# Tests
1. What unit/integration tests cover your change? What autoQA tests cover your change?
1. What tests did you perform that validates your changed worked?

# QA
1. What does QA need to do to validate your changes?
1. What areas to they need to test for regressions?

PR #721, opened 2024-06-13
[TS migration] Migrate ExpensiMark to Typescript

blazejkustra:
This PR aims to migrate `ExpensiMark` to Typescript! 
- `'quote'` rule was adjusted as it was unexpectedly using a boolean flag as part of `replacement` function (`shouldKeepRawInput`). Instead I implemented `rawInputReplacement` for this rule
- `extras` object has to be moved from being the last argument to be the first argument of every `replacement` and `rawInputReplacement` functions. 

<details>
<summary>Context</summary>
`ReplacementFn` was invalidly defined in `.d.ts` file, but TS server don't check `.d.ts` and therefore it wasn't detected eariler. Now once I moved it to a `.ts` file the issue was discovered and there was no other approach that worked.

```ts
// before:
type Replacement = (...matches: string[], extras?: Extras) => string; // TS ERROR

// after
type ReplacementFn = (extras: Extras, ...matches: string[]) => string;
```
</details>

~~PR is on hold as it's waiting for this [PR](https://github.com/Expensify/expensify-common/pull/718) to be merged first.~~

### Fixed Issues
$ https://github.com/Expensify/App/issues/42247

# Tests
- Test all possible live-markdown features (quotes, headings, code, inline code etc)

```
``` - code block
``` 
`x` - code inline
`[]()` - link
`> text` - quote
`# text` - heading

Also, nested quotes, and other edge cases

<details>

E/App:
image

live-markdown:
image

</details>


lib/ExpensiMark.ts
```diff
@@ -260,23 +324,43 @@ export default class ExpensiMark {
...
-                        if (shouldKeepRawInput) {
-                            isStartingWithSpace = !!g2;
-                            return '';
-                        }
-                        return match;
+                    const handleMatch = (_match: string, g2: string) => {
+                        isStartingWithSpace = !!g2;
+                        return '';
                     };
                     const textToReplace = g1.replace(/^&gt;( )?/gm, handleMatch);
```
Skalakid, line 365: Why now we aren't returning `match` in handleMatch function if `isStartingWithSpace` is false?

lib/ExpensiMark.ts
```diff
@@ -604,40 +686,43 @@ export default class ExpensiMark {
...
-         * @type {Object[]}
          */
         this.whitespaceRulesToDisable = ['newline', 'replacepre', 'replacebr', 'replaceh1br'];
 
         /**
          * The list of rules that have to be applied when shouldKeepWhitespace flag is true.
-         * @param {Object} rule - The rule to check.
-         * @returns {boolean} Returns true if the rule should be applied, otherwise false.
+         * @param rule - The rule to check.
+         * @returns Returns true if the rule should be applied, otherwise false.
```
Skalakid, line 695:
```suggestion
         * @returns true if the rule should be applied, otherwise false.
```

lib/ExpensiMark.ts
```diff
@@ -1079,46 +1131,40 @@ export default class ExpensiMark {
     }
 
     /**
-     * @param {String} comment
-     * @returns {Array} or undefined if exception occurs when executing regex matching
+     * @returns or undefined if exception occurs when executing regex matching
```
Skalakid, line 1134:
```suggestion
     * @returns array or undefined if exception occurs when executing regex matching
```

lib/ExpensiMark.ts
```diff
@@ -1079,46 +1131,40 @@ export default class ExpensiMark {
...
         const linksInNew = this.extractLinksInMarkdownComment(newComment);
         return linksInOld === undefined || linksInNew === undefined ? [] : linksInOld.filter((link) => !linksInNew.includes(link));
     }
 
     /**
      * Escapes the content of an HTML attribute value
-     * @param {String} content - string content that possible contains HTML
-     * @returns {String} - original MD content escaped for use in HTML attribute value
+     * @param content - string content that possible contains HTML
+     * @returns - original MD content escaped for use in HTML attribute value
```
Skalakid, line 1165:
```suggestion
     * @returns original MD content escaped for use in HTML attribute value
```

Also what is MD? Markdown?

lib/ExpensiMark.ts
```diff
@@ -115,6 +176,9 @@ export default class ExpensiMark {
                 name: 'heading1',
                 process: (textToProcess, replacement, shouldKeepRawInput = false) => {
                     const regexp = shouldKeepRawInput ? /^# ( *(?! )(?:(?!<pre>|\n|\r\n).)+)/gm : /^# +(?! )((?:(?!<pre>|\n|\r\n).)+)/gm;
+                    if (typeof replacement === 'function') {
+                        return textToProcess.replace(regexp, (...args) => replacement(EXTRAS_DEFAULT, ...args));
+                    }
                     return textToProcess.replace(regexp, replacement);
```
Skalakid, line 182: I see that in 5 places we are doing something similar. What about moving this logic to separate function?
For example:

```ts
function replaceText(text: string, regexp: RegExp, extras, replacement: Replacement) {
    if (typeof replacement === 'function') {
        return text.replace(regexp, (...args) => replacement(extras, ...args));
    }
    return text.replace(regexp, replacement);
}
```

Skalakid (commented):
LGTM, I left some small comments

lib/ExpensiMark.ts
blazejkustra: Previously there was no `rawInputReplacement` function. What I did is to split previous logic into two functions (`replacement` and `rawInputReplacement`). `match` is returned but only in `replacement` function above

lib/ExpensiMark.ts
```diff
@@ -536,16 +614,15 @@ export default class ExpensiMark {
...
-                 * @param {string} videoName - the fourth capture group will be the video file name (the text between opening and closing video tags)
-                 * @param {any[]} args The rest of the arguments. args[args.length-1] will the `extras` object
-                 * @returns {string} Returns the markdown video tag
+                 * @param match The full match
+                 * @param g1 {string} The first capture group
+                 * @param videoSource - the second capture group - video source (video URL)
+                 * @param videoAttrs - the third capture group - video attributes (data-expensify-width, data-expensify-height, etc...)
+                 * @param videoName - the fourth capture group will be the video file name (the text between opening and closing video tags)
+                 * @param args The rest of the arguments. args[args.length-1] will the `extras` object
+                 * @returns Returns the markdown video tag
```
Skalakid, line 623:
```suggestion
                 * @returns The markdown video tag
```

lib/ExpensiMark.ts
```diff
@@ -1095,19 +1126,17 @@ export default class ExpensiMark {
...
         return textToCheck;
     }
 
     /**
      * Check if the input text includes only the open or the close tag of an element.
      *
-     * @param {String} textToCheck - Text to check
-     *
-     * @returns {Boolean}
+     * @param textToCheck - Text to check
```
Skalakid, line 1137: I think we don't need prop explanation here
```suggestion
```

Skalakid (approved):
LGTM, left 2 small suggestions

lib/ExpensiMark.ts
```diff
@@ -16,33 +54,59 @@ const SLACK_SPAN_NEW_LINE_TAG = '<span class="c-mrkdwn__br" data-stringify-type=
...
         ExpensiMark.Log = logger;
     }
 
+    rules: Rule[];
+
+    htmlToMarkdownRules: RuleWithRegex[];
+
+    htmlToTextRules: RuleWithRegex[];
+
+    whitespaceRulesToDisable = ['newline', 'replacepre', 'replacebr', 'replaceh1br'];
```
fabioh8010, line 77: What about adding descriptions for these properties too?

fabioh8010 (commented):
@blazejkustra Why did you have to do that?

> `extras` object has to be moved from being the last argument to be the first argument of every `replacement` and `rawInputReplacement` functions.

Ideally extras should be a the end (the other functions are still keeping it at the end).

blazejkustra:
> @blazejkustra Why did you have to do that?
> 
> > `extras` object has to be moved from being the last argument to be the first argument of every `replacement` and `rawInputReplacement` functions.
> 
> Ideally extras should be a the end (the other functions are still keeping it at the end).

@fabioh8010 Ideally yes, but it's not possible to have an argument after "spread argument":
```ts
type Replacement = (...matches: string[], extras?: Extras) => string; // TS ERROR
```

This throws an error: "A rest parameter must be last in a parameter list.". That is why I had to move it to be first

lib/ExpensiMark.ts
```diff
@@ -134,19 +210,17 @@ export default class ExpensiMark {
                 name: 'video',
                 regex: MARKDOWN_VIDEO_REGEX,
                 /**
-                 * @param {string} match
-                 * @param {string} videoName - The first capture group - video name
-                 * @param {string} videoSource - The second capture group - video URL
-                 * @param {any[]} args - The rest capture groups and `extras` object. args[args.length-1] will the `extras` object
-                 * @return {string} Returns the HTML video tag
+                 * @param match
```
BartoszGrajdek, line 213:
```suggestion

```

lib/ExpensiMark.ts
```diff
@@ -134,19 +210,17 @@ export default class ExpensiMark {
...
+                 * @param videoName - The first capture group - video name
+                 * @param videoSource - The second capture group - video URL
+                 * @param args - The rest capture groups and `extras` object. args[args.length-1] will the `extras` object
```
BartoszGrajdek, line 216: This should probably change

BartoszGrajdek (approved):
Overall looks good, I'll test it with the `react-native-live-markdown` lib and let you know how it works out! 🙌🏻

lib/ExpensiMark.ts
BartoszGrajdek: The `args` param ^

fabioh8010 (approved):
LGTM, let's do extensive testing to make sure everything is working

blazejkustra:
@srikarparsi All yours! Already tested in E/App and `react-native-live-markdown` repos

srikarparsi (approved):
nice! this looks good to me

srikarparsi:
We also need to bump the version in react-native-live-markdown and in App. I can get PRs out for those.

PR #730, opened 2024-06-25
capture newline as the code fence content

bernhardoj:
### Fixed Issues
$ https://github.com/Expensify/App/issues/43750

# Tests
Updated existing unit test for codefence.

To test this in App,
- open live markdown repo, apply the expensi-mark and live-markdown fix
- run `cd parser && npm run build`
- copy the `parser/react-native-live-markdown-parser` to App `node_modules/@expensify/react-native-live-markdown/lib/parser/react-native-live-markdown-parser.js`.

1. Submit an expense with code block as description
2. Open the transaction thread
3. Open the description
4. Verify there is no console error of react-native-live-markdown

<details>
<summary>Web</summary>

https://github.com/Expensify/expensify-common/assets/50919443/f8ed1153-cc6a-4b7a-9608-1de7b1205d90

</details>

<details>
<summary>Desktop</summary>

https://github.com/Expensify/expensify-common/assets/50919443/b5b2243a-5200-4646-b8c0-b35af6c89bd6

</details>

<details>
<summary>iOS mWeb</summary>

https://github.com/Expensify/expensify-common/assets/50919443/bc9d57f2-9748-4b0c-b844-f69617874a5e

</details>

<details>
<summary>iOS Native</summary>

https://github.com/Expensify/expensify-common/assets/50919443/ea4d9b2c-7f75-4053-a27a-5f44de096d81

</details>

<details>
<summary>Android mWeb</summary>

https://github.com/Expensify/expensify-common/assets/50919443/cffbdd8a-6dac-4339-b264-0d1338aa563c

</details>

<details>
<summary>Android Native</summary>

https://github.com/Expensify/expensify-common/assets/50919443/56c2d920-08ca-4353-862a-e716804e3e3f

</details>

# QA
Same as Tests above


bernhardoj:
@BartoszGrajdek here is the PR

BartoszGrajdek:
Testing today!

BartoszGrajdek (commented):
Looks good to me - please attach some test videos for other platforms, because these changes affect all of them 🙏🏻

__tests__/ExpensiMark-HTML-test.js
```diff
@@ -2220,7 +2220,7 @@ describe('room mentions', () => {
 
     test('room mention inside code block should not be rendered', () => {
         const testString = '```\n#room\n```';
-        const resultString = '<pre>#room<br /></pre>';
+        const resultString = '<pre><br />#room<br /></pre>';
```
s77rt, line 2223: I believe this is unwanted. HTML BR collapsing works for `<br /></pre>` but not for `<pre><br />` https://codesandbox.io/s/silly-brattain-dz3rrw

s77rt (changes requested):

__tests__/ExpensiMark-HTML-test.js
bernhardoj: Interesting. The App PreRenderer works fine though, but I think we may want to avoid this. Maybe we can put the new line info on the <pre> tag like:
```
<pre data-newline="\r\n">
```

and live markdown can get the new line from `data-newline`.

s77rt: That would work but it won't work if we have mixed newlines. Technically the live markdown works correctly and the console error is a false positive. Maybe we can ignore the error if the difference between the original text and the reconstructed one is only the new lines? Though ultimately we want the reconstructed message to be the same

bernhardoj: > That would work but it won't work if we have mixed newlines

Hmm, you're right.

> Maybe we can ignore the error if the difference between the original text and the reconstructed one is only the new lines?

I think we can replace all the new lines to \n only when doing the comparison. This way, we don't alter the user input at all, but we need @BartoszGrajdek thoughts on this one.

BartoszGrajdek: What if we added the line break just in `rawInputReplacement`? The returned value of it is coming only into `react-native-live-markdown` input, so it won't affect how we treat code blocks in the chat itself, just the input. 

We could just treat the linebreak before code block content as a separate group and use it only in `rawInputReplacement`.

BartoszGrajdek: Something like this:
```
replacement: (_extras, _match, _g1, _g2, textWithinFences) => {
    const group = textWithinFences.replace(/(?:(?![\n\r])\s)/g, '&#32;');
    return `<pre>${group}</pre>`;
},
rawInputReplacement: (_extras, _match, _g1, g2, textWithinFences) => {
    const group = textWithinFences.replace(/(?:(?![\n\r])\s)/g, '&#32;').replace(/<emoji>|<\/emoji>/g, '');
    return `<pre>${g2}${group}</pre>`;
},

Comment (bernhardoj) [reply __tests__/ExpensiMark-HTML-test.js:2223]: Yeah, that works.

@s77rt do you agree with the above suggestion?

Comment (s77rt) [reply __tests__/ExpensiMark-HTML-test.js:2223]: Sounds good to me but can you please test this against `react-native-live-markdown` and verify that we don't get extra newlines

Comment (bernhardoj) [reply __tests__/ExpensiMark-HTML-test.js:2223]: Yeah, I don't see any extra newlines
image

Comment (s77rt) [reply __tests__/ExpensiMark-HTML-test.js:2223]: Let's go with that then 👍

Comment (bernhardoj) [reply __tests__/ExpensiMark-HTML-test.js]: Updated!

File: lib/ExpensiMark.ts [hunk 2]
```diff
@@ -129,21 +129,21 @@ export default class ExpensiMark {
                 name: 'codeFence',
 
                 // &#x60; is a backtick symbol we are matching on three of them before then after a new line character
-                regex: /(&#x60;&#x60;&#x60;(?:\r\n|\n))((?:\s*?(?!(?:\r\n|\n)?&#x60;&#x60;&#x60;(?!&#x60;))[\S])+\s*?(?:\r\n|\n))(&#x60;&#x60;&#x60;)/g,
+                regex: /(&#x60;&#x60;&#x60;(\r\n|\n))((?:\s*?(?!(?:\r\n|\n)?&#x60;&#x60;&#x60;(?!&#x60;))[\S])+\s*?(?:\r\n|\n))(&#x60;&#x60;&#x60;)/g,
```
s77rt, line 132: NAB. This has 4 capturing groups but we only need 2. Can you make the other two non-capturing groups?

lib/ExpensiMark.ts
```diff
@@ -129,21 +129,21 @@ export default class ExpensiMark {
...
                 // with the new lines here since they need to be converted into <br>. And we don't
                 // want to do this anywhere else since that would break HTML.
                 // &nbsp; will create styling issues so use &#32;
-                replacement: (_extras, _match, _g1, textWithinFences) => {
+                replacement: (_extras, _match, _g1, _g2, textWithinFences) => {
                     const group = textWithinFences.replace(/(?:(?![\n\r])\s)/g, '&#32;');
                     return `<pre>${group}</pre>`;
                 },
-                rawInputReplacement: (_extras, _match, _g1, textWithinFences) => {
+                rawInputReplacement: (_extras, _match, _g1, g2, textWithinFences) => {
```
s77rt, line 144: Rename the group variable that has the newline to something more fitting / explaining what the groups holds e.g. `newLineCharacter`

s77rt (changes requested):

lib/ExpensiMark.ts
bernhardoj: I think there is 6 groups in total including the negative lookahead and we use 3, but I will leave it as it is.

lib/ExpensiMark.ts
bernhardoj: Right, updated!

s77rt (approved):

BartoszGrajdek:
Hi! Code changes look good to me, I'll just try testing it inside `react-native-live-markdown` lib tomorrow morning before approving to ensure no regressions are introduced. 

I'm currently working on an issue with higher priority, hence the delay 🙏🏻

BartoszGrajdek (approved):
LGTM - Works well in `react-native-live-markdown` ✅

bernhardoj:
Will prepare the App PR tomorrow!

PR #746, opened 2024-07-08
Rename `escape`/`unescape` to `escapeText`/`unescapeText`

tomekzaw:
This PR renames `escape` and `unescape` functions in `Utils` module to `escapeText` and `unescapeText`, accordingly.

The reason for this change is that, although deprecated, both `escape` and `unescape` are built-in JavaScript functions:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape

Hence, these names might be confusing for other third-party libraries that rely on this rule, e.g. Reanimated Babel plugin.

These changes are needed for `ExpensiMark` to work properly in worklets in `react-native-live-markdown`:
* https://github.com/Expensify/react-native-live-markdown/issues/317

### Fixed Issues
$ GH_LINK

# Tests

# QA

### DIFF
--- a/__tests__/ExpensiMark-test.js
+++ b/__tests__/ExpensiMark-test.js
@@ -25,7 +25,7 @@ test('Test with regex Maximum regex stack depth reached error', () => {
         throw new Error('Maximum regex stack depth reached');
     });
     expensiMarkParser.modifyTextForUrlLinks = modifyTextForUrlLinksMock;
-    expect(expensiMarkParser.replace(testString)).toBe(Utils.escape(testString));
+    expect(expensiMarkParser.replace(testString)).toBe(Utils.escapeText(testString));
     expect(modifyTextForUrlLinksMock).toHaveBeenCalledTimes(1);
 
     // Mock method extractLinksInMarkdownComment to let it return undefined to test try/catch of method ExpensiMark.extractLinksInMarkdownComment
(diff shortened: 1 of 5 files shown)


BartoszGrajdek (approved):

roryabraham (approved):
hmmm seems ok, but also I wonder if the Reanimated Babel plugin should be able to distinguish between a global function call and a function call in a module namespace.

Not really important enough that I want to block this PR, so I'm going to proceed as-is

roryabraham:
got some test failures though

roryabraham:
Marking as WIP until tests are ✅

tomekzaw:
> I wonder if the Reanimated Babel plugin should be able to distinguish between a global function call and a function call in a module namespace.

I think this should be technically possible but I believe @tjzel can provide more details on that.

Currently we just have a list of the names of the identifiers that never get captured: https://github.com/software-mansion/react-native-reanimated/blob/77644ca17d54b20a5547e7a1be8ec279cd2b5ded/packages/react-native-reanimated/plugin/src/globals.ts#L2-L116

I'll write it down on our TODO list.

> got some test failures though

My bad, I forgot to update the function names in the exports. Should be fixed now

roryabraham (approved):

roryabraham:
Whoops, can't merge this with unsigned commits. Please rebase and resign commits.

tomekzaw:
@roryabraham This is my first contribution to `expensify-common` so I forgot to enable signing commits. Sorry for inconvenience. Pushed a new signed commit with all changes.

tjzel:
@roryabraham It seems possible to distinguish between such calls, I'll try to explore that direction in the future.

PR #788, opened 2024-08-21
Allow codefence regex to accept any characters after first three backticks

tsa321:
### Fixed Issues
$ https://github.com/Expensify/App/issues/45088

# Tests
I have add unit tests for this issue.
Update the related regex in the expensify-common in node module and send the test messages from composer of the App.
Test example:
````
```   
test
```
````

````
```  abc 
test
```
````

The code fence/ code block must be displayed properly.

Verify the displayed message is displayed correctly.

# QA
Same as test above
Code fence related regression tests.

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -454,6 +454,12 @@ test('Test code fencing with spaces and new lines', () => {
     codeFenceExample = '```\r\n# test\r\n```';
     expect(parser.replace(codeFenceExample)).toBe('<pre>#&#32;test<br /></pre>');
 
+    codeFenceExample = '```   \r\n# test\r\n```';
+    expect(parser.replace(codeFenceExample)).toBe('<pre>#&#32;test<br /></pre>');
+
+    codeFenceExample = '```test\r\n# test\r\n```';
+    expect(parser.replace(codeFenceExample)).toBe('<pre>#&#32;test<br /></pre>');
+
     codeFenceExample = '```\r\n\r\n# test\r\n\r\n```';
     expect(parser.replace(codeFenceExample)).toBe('<pre><br />#&#32;test<br /><br /></pre>');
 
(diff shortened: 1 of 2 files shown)


akinwale:
> @akinwale Just to confirm when sending message for example:
> 
> ```
> ```aaa
> bbb
> ```
> ```
> 
> The displayed message is:
> 
> ```
> bbb
> ```
> 
> The `aaa` text is removed. When editing the message, the `aaa` text no longer appears. It looks like:
> 
> ```
> ```
> bbb
> ```
> ```
> 
> This is expected behavior right?

I don't think this is expected behaviour. The characters within the codefence should be included instead of being stripped out.

akinwale:
> @akinwale so the expected result would be:
> 
> ```
> aaa
> bbb
> ```
> 
> Right?

Yes, this is correct.

deetergp:
> The characters within the codefence should be included instead of being stripped out.

I will volley here and say that yes, this is the expected behavior for every Markdown variant I have ever come across. If you want your text to appear in your codefence, you need to make sure it's after the line containing the triple backticks. In some variants the area after the backticks is reserved for specifying what flavor of context highlighting you want for you code.

deetergp:
Yeah, I think allowing anything after the three backticks is fine, but expected behavior should be that they are not rendered in the code block.

deetergp:
@tsa321 Let's not block on that. Right now we don't really support any kind of context highlighting, so there's no real reason for text to be there anyway.

akinwale:
> kindly bump for review @akinwale. Or maybe @AndrewGable Could you review this instead?

Reviewing this today.

akinwale (approved):
LGTM.

akinwale:
@tsa321 Looks good, thanks. Will wait for internal review from @deetergp or @AndrewGable and then you can create the corresponding PR in the Expensify App repo.

AndrewGable (approved):

PR #847, opened 2025-04-17
Remove undefined methods and replace them with native ones

youssef-lr:
### Fixed Issues
$ https://github.com/Expensify/Expensify/issues/489667

# Tests
1. Verify `npm run test` works in Web-E.
2. Verify loading various pages works fine.

# QA
N/A, will be test in Web-E.

### DIFF
--- a/lib/jquery.expensifyIframify.js
+++ b/lib/jquery.expensifyIframify.js
@@ -351,7 +351,7 @@ export default {
                         const object = obj;
                         delete object[name];
                     };
-                    _.each(eventHandlers, removeEventHandler);
+                    eventHandlers.forEach(removeEventHandler);
                 }
             },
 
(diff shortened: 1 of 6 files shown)


lib/jquery.expensifyIframify.js
```diff
@@ -217,7 +217,7 @@ export default {
                 return;
             }
 
-            if (!event.data || !_.isString(event.data)) {
+            if (!event.data || typeof event.data !== 'string') {
```
cead22, line 220: According to grok these aren't 100% equivalent
```suggestion
            if (!event.data || (typeof event.data !== 'string' && !(event.data instanceof String))) {
```
image

cead22 (approved):

PR #850, opened 2025-04-22
Fix quoted multiline links

Skalakid:
This PR fixes an issues with parsing quoted multiline link which causes errors in the Live Markdown Input.

Now it looks like this:
Screenshot 2025-04-22 at 15 29 47

### Fixed Issues

$ GH_LINK

# Tests

# QA

### DIFF
--- a/__tests__/ExpensiMark-HTML-test.js
+++ b/__tests__/ExpensiMark-HTML-test.js
@@ -1751,6 +1751,19 @@ test('Test link with header before the alias multiline text part', () => {
     expect(parser.replace(testString)).toBe(resultString);
 });
 
+
+test('Test markdown replacement for quoted multiline links', () => {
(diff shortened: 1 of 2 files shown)


JS00001:
Hmm, I'm getting 
image
Is this correct?

Skalakid:
Changing it to draft so I can re-verify what the desired outcome is

Skalakid:
Found out that email labels can be multiline, even inside the quote. So I will update my solution to make links consistent with it, since the syntax is similar

Screenshot 2025-04-23 at 15 18 25

Skalakid:
Okay, everything should work now. I changed the expected result, so keep it in mind while testing :D

JS00001 (approved):
Looks good :shipit:

PR #852, opened 2025-05-02
feat: add maxBodyLength parameter to htmlToMarkdown

yellowtailfan:
Supporting change for fixing https://github.com/Expensify/App/issues/57241.

# Tests
* Added "Test long HTML header" to `__tests__/ExpensiMark-Markdown-test.js`
* Tested on all platforms from App code modified to use new `maxBodyLength` parameter.
* Tests include HTML clipboard paste on platforms where available.
* Tests included very large clipboard pastes to check for performance issues. There was some lag but it was similar both before and after this change.
* Inspected all calls to `htmlToMarkdown()` from Expensify repo and none will be affected by the additional parameter.

# QA
* Run unit tests.
* HTML to Markdown conversion.

### DIFF
--- a/lib/ExpensiMark.ts
+++ b/lib/ExpensiMark.ts
@@ -1234,8 +1234,14 @@ export default class ExpensiMark {
             generatedMarkdown = parseBodyTag[2];
         }
         if (maxBodyLength > 0) {
-            // Truncate the text after extracting the body tag, as some HTML sources (such as Microsoft Word)
-            // have headers larger than the typical maxLength of 10K even for a small body text.
+            /*
+             * Some HTML sources (such as Microsoft Word) have headers larger than the typical maxLength of
+             * 10K even for a small body text. So the text is truncated after extracting the body element, to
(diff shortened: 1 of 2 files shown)


yellowtailfan:

lib/ExpensiMark.ts
```diff
@@ -1233,6 +1233,17 @@ export default class ExpensiMark {
         if (parseBodyTag) {
             generatedMarkdown = parseBodyTag[2];
         }
+        if (maxBodyLength > 0) {
```
jjcoffee, line 1236: Coming from https://github.com/Expensify/App/issues/63711. We should have decoded the HTML first before truncation, otherwise decoding can fail if there isn't enough content to decode. More details [here](https://github.com/Expensify/App/issues/63711#issuecomment-2956092514).

yellowtailfan: Here is a more detailed comment on the performance issues around truncation before or after parsing: https://github.com/Expensify/App/issues/57241#issuecomment-2744969896

PR #869, opened 2025-07-09
Fix fastMerge worklet

Skalakid:
We've tried updating `expensify-common` to the latest version (2.0.144) in the `react-native-live-markdown` repo, but encountered crashes due to invalid fastMerge file workletization caused by https://github.com/Expensify/expensify-common/pull/855. Because in this file we can find double-function recursion and worklets aren't hoisted, it results in a crash with an error: `Cannot access 'fastMerge' before initialization`.

In this PR, I'm breaking double function recursion and moving the `mergeObject` code inside the `fastMerge` function. It shouldn't affect anything since `mergeObject` isn't exported, only `fastMerge` was using it.

### Fixed Issues
$ https://github.com/Expensify/App/issues/39518

# Tests
1. Verify if E/App isn't crashing in LHN with this change
2. Verify if the `live-markdown` example app isn't crashing too with this change

# QA
Same as Tests

### DIFF
--- a/lib/fastMerge.ts
+++ b/lib/fastMerge.ts
@@ -13,21 +13,28 @@ function isMergeableObject(value: unknown): value is Record<string, unknown> {
 }
 
 /**
- * Merges the source object into the target object.
- * @param target - The target object.
- * @param source - The source object.
- * @param shouldRemoveNestedNulls - If true, null object values will be removed.
- * @returns - The merged object.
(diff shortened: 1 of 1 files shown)


adderly:
This is needed for using the latest versions of markdown library.

mollfpr (approved):
LGTM!

adderly:
Thank you

PR #870, opened 2025-07-09
remove truncation when convert html to markdown

nkdengineer:
### Fixed Issues
$ https://github.com/Expensify/App/issues/63711

# Tests

# QA

### Result

https://github.com/user-attachments/assets/07d6019f-ff64-4d4f-8c4c-6b8837fed160


justinpersaud (commented):
@nkdengineer your proposal mentioned we would

>Truncate the final result

Is that no longer necessary?

Also, are there any other tests we should do with regards to pasting now that truncation is being removed?

PR #875, opened 2025-08-04
Add more constants since web-s has a constant file deleted

techievivek:
CC @AndrewGable @madmax330 
Coming from discussion here https://expensify.slack.com/archives/C03SSAQ7P/p1754285585977329, we observed that a few constants that were removed as part of deleting `00constant.js` file exists nowhere in our codebase and resulted in a few regression so through this PR we try to backfill the missing constants.

### Fixed Issues
N/A

# Tests
1. Login to OldDot.
2. Navigate to `Settings > Wallet`.
3. Click on `Import Card/Bank` and select any option among first 6.
4. Confirm `Log In` page is shown.

# QA
Same as tests.

### DIFF
--- a/lib/CONST.ts
+++ b/lib/CONST.ts
@@ -32,19 +32,6 @@ const CONST = {
     URL_TO_SECURE_WEBSITE: '',
 
     /**
-     * This is the maximum number of transactions that can be exported to quickbooks
-     *
-     * @type {Number}
-     */
-    QUICKBOOKS_EXPORT_LIMIT: 1500,
-
-    /**
-     * UNUSED, BY KEPT FOR "STORAGE"
-     * This key is only valid for URLs in the https://www.expensify.com/ realm.
-     */
-    GOOGLE_API_KEY: 'ABQIAAAAjtisvqrcGboNGaMCOuZ0ohQhhEh1oy_9rMnViic8Bkkf7v7oORR3gYYz-G4SlC7hy72TbmOpwIHAtg',
-
-    /**
      * IRS remimbursement rate for mileage
      * WARNING ! UPDATE THE PHP CONSTANT VERSION WHEN UPDATING THIS ONE
      *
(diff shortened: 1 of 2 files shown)


lib/CONST.ts
```diff
@@ -25,6 +25,25 @@ const CONST = {
...
+     * @type {String}
+     */
+    URL_TO_SECURE_WEBSITE: '',
+
+    /**
+     * This is the maximum number of transactions that can be exported to quickbooks
+     *
+     * @type {Number}
+     */
+    QUICKBOOKS_EXPORT_LIMIT: 1500,
```
MonilBhavsar, line 39: Unused, so can me removed?
```suggestion
```

lib/CONST.ts
```diff
@@ -485,6 +512,13 @@ const CONST = {
          */
         MOMENT_DATE_TIME: 'YYYY-MM-DD HH:mm',
 
+        /**
+         * This is a typical format of the date plus the time and seconds
+         *
+         * @type {String}
+         */
+        MOMENT_DATE_TIME_WITH_SECONDS: 'YYYY-MM-DD HH:mm:ss',
+
```
MonilBhavsar, line 520: Same, unused?
```suggestion
```

lib/CONST.ts
```diff
@@ -572,6 +585,23 @@ const CONST = {
...
+    // see https://stackoverflow.com/c/expensify/questions/1289/1290#1290 for more about this.
+    // If a bank isn't on this list, we default to 90 days
+    BANK_IMPORT_DAY_LIMITS: {
+        'americanexpress.com': 30,
+        'oauth.bankofamerica.com': 30,
+        'oauth.capitalone.com': 90,
+        'oauth.chase.com': 90,
+        'discover.com': 30,
+        'admin.pexcard.com': 30,
+        svbdirect: 75,
```
MonilBhavsar, line 598: Should this be a string?

```suggestion
        'svbdirect': 75,
```

madmax330 (dismissed):

PR #914, opened 2026-05-15
Move COUNTRY_ZIP_REGEX_DATA and GENERIC_ZIP_CODE_REGEX from App into expensify-common

jasperhuangg:
Move `COUNTRY_ZIP_REGEX_DATA` (the per-country zip-code regex/sample table) and `GENERIC_ZIP_CODE_REGEX` (the catch-all fallback) out of NewDot's local `CONST` ([App/src/CONST/index.ts:5263-6082](https://github.com/Expensify/App/blob/main/src/CONST/index.ts#L5263-L6082)) and into `expensify-common` so both NewDot and OldDot can use the same source of truth.

Today the OldDot card-replacement form ([requestNewCard.jsx](https://github.com/Expensify/Web-Expensify/blob/main/site/app/dialogs/cards/requestNewCard.jsx)) uses a hardcoded US-only zip regex with no validation for other countries, while NewDot has a much richer per-country regex table. This PR is the first of three that consolidates both apps onto the shared data; follow-ups will land in `Expensify/App` and `Expensify/Web-Expensify`.

No behavior change in this repo on its own — the constants are added but nothing in `expensify-common` consumes them yet. Bumped the package version to `2.0.179` so the App and Web-Expensify PRs can pin it.

### Fixed Issues
part of https://github.com/Expensify/Expensify/issues/635650

# Tests
1. `npm run typecheck`, `npm run build`, `npm run lint`, and `npx jest --watchman=false` (418 tests) all pass on this branch.
2. Verified `dist/CONST.js` and `dist/CONST.d.ts` after `npm run build` contain the new `COUNTRY_ZIP_REGEX_DATA` and `GENERIC_ZIP_CODE_REGEX` entries (lines 912 and 1727 in `CONST.js`).

# QA
1. Once published to npm, install in App and Web-Expensify and confirm `CONST.COUNTRY_ZIP_REGEX_DATA.US.regex.test('12345') === true` and `CONST.GENERIC_ZIP_CODE_REGEX` is a `RegExp`.
2. No runtime regression risk in this PR alone; the consumer-side PRs (App and Web-Expensify) carry the full QA plans.

Made with [Cursor](https://cursor.com)


rafecolton (approved):
LGTM

PR #922, opened 2026-06-19
Remove CLI from barrel; export as TypeScript subpath

roryabraham:
## Problem

Adding `CLI` to the main barrel in 2.0.184 caused every `import {Str} from 'expensify-common'` to eagerly load `CLI.js`, which pulls in Node's `readline`. That broke Metro/HybridApp builds in App.

CLI is Node-only (scripts, ts-node, tsx). It should never be on the compiled CJS barrel.

## Solution

- Remove `CLI` from `lib/index.ts` — the barrel no longer references it.
- Exclude `lib/CLI.ts` from `tsconfig.build.json` so `tsc` does not emit `dist/CLI.js`.
- Add a `package.json` `exports` subpath: `expensify-common/CLI` → `lib/CLI.ts` (source TypeScript, consumed by TS-native runtimes).
- Publish `lib/SafeString.ts` alongside `lib/CLI.ts` because CLI imports it at the source level.
- Add `exports` map for the main entry (`.`) with explicit `require`/`types` conditions.

No lazy-loading proxies, no ESM facade files, no build-script dist cleanup.

## Consumer migration

```ts
// Before
import {CLI} from 'expensify-common';

// After
import CLI from 'expensify-common/CLI';
```

Downstream follow-ups (separate PRs):
- **App** — update script imports; can drop Metro CLI stub once on this version
- **GitHub-Actions** — simplify `scripts/libs/CLI.ts` to import from `expensify-common/CLI`

## Test plan

- [x] `npm run build` — `dist/CLI.js` is not produced
- [x] `npm run typecheck` — `CLI.ts` still typechecked via root tsconfig
- [x] `npm test`
- [x] `npm run lint`
- [x] Merged `main`; version left unchanged (publish workflow bumps post-merge)


chuckdries (approved):
🤩

PR #923, opened 2026-06-26
Compile CLI subpath to dist instead of shipping TS source

roryabraham:
Follow-up to https://github.com/Expensify/expensify-common/pull/922.

## Problem

The `expensify-common/CLI` subpath pointed at raw `lib/CLI.ts` source. That caused two consumer issues:

- **tsx / ESM**: static `import CLI from 'expensify-common/CLI'` resolved to a module namespace object instead of the constructor, requiring `.default` unwrap hacks
- **Node 24 native TS**: `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING` because Node refuses to strip types from `node_modules`

## Solution

- Compile `CLI.ts` to `dist/CLI.js` (CJS) via a dedicated `tsconfig.cli.cjs.json` build step
- Add a thin ESM entry at `dist/esm/CLI.js` that re-exports the compiled default for the `import` condition
- Update `package.json` exports:
  - `import` → `dist/esm/CLI.js`
  - `require` → `dist/CLI.js`
  - `types` → `dist/CLI.d.ts`
- Stop publishing `lib/CLI.ts` / `lib/SafeString.ts` in the npm tarball (both remain in repo for development)
- Expand `package-exports-test.js` to verify CJS/ESM default export interop

CLI remains excluded from the main CJS barrel so Metro/HybridApp imports of `expensify-common` are unaffected.

## Downstream follow-ups (separate PRs)

- **GitHub-Actions** — simplify `eventContext.ts` to bare `import CLI from 'expensify-common/CLI'` and drop `type-fest` devDep + `.default` unwrap once on the new version

## Tests

- [x] `npm run build`
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm test` (472 tests)
- [x] Verified bare `import CLI from 'expensify-common/CLI'` works with Node 24 ESM and tsx

Made with [Cursor](https://cursor.com)


package.json
```diff
@@ -14,22 +14,22 @@
       "default": "./dist/index.js"
     },
     "./CLI": {
-      "types": "./lib/CLI.ts",
-      "default": "./lib/CLI.ts"
+      "types": "./dist/CLI.d.ts",
+      "import": "./dist/esm/CLI.js",
+      "require": "./dist/CLI.js",
```
chuckdries, line 19: Not a blocker, but if we don't actually `require()` the CLI class anywhere, we can drop the CJS bundle entirely

chuckdries (approved):

PR #926, opened 2026-07-06
Implement truncate mechanism for Log messages

fabioh8010:
### Details

Big logs were being sent to the server and rejected because the server flags any single serialized log line larger than **1,048,576 bytes (1 MiB = 2²⁰)** — see https://github.com/Expensify/App/issues/95227 (e.g. `[App] LogAPI - Massive log message given`). In practice these originate in `react-native-onyx`, which bakes large data straight into the log **message** string (e.g. `set called for key: loginToAccountIDMap properties: <huge comma-separated key list>`), and in E/App call sites that pass large objects as **parameters**.

This PR adds a hard byte-size safeguard inside the Logging framework itself, so every consumer (Onyx, App, other platforms) is protected without changing individual call sites. The limit is enforced on the **serialized line** — exactly what the server measures — so it accounts for `message`, `parameters`, metadata (`timestamp`, `email`) and JSON-escaping overhead together.

**What changed (`lib/Logger.ts`):**
- New `MAX_LOG_LINE_BYTES = 1_000_000` constant — capped below the server's 1,048,576 to leave headroom for JSON escaping and the surrounding line fields.
- Enforcement happens once, at flush time in `logToServer()`, via `serializeLineWithinByteLimit(line, maxSize)`:
  - It measures the **serialized** line size (`JSON.stringify`), so JSON escaping (quotes/backslashes → 2×, control chars & lone surrogates → `\uXXXX` 6×) is fully accounted for.
  - If the line is over the limit even with an empty message, the bulk is in `parameters` (structured data we can't safely truncate mid-JSON), so it's replaced with a `{truncated: true, originalByteSize: N}` marker and the human-readable message is kept.
  - The `message` is then truncated with `truncateMessageToFitLine()` — a **single pass** that sizes each code point by its exact JSON-escaped byte length (`jsonEscapedByteSize`), never splits a multi-byte character, and appends a `...[truncated N bytes]` marker (N = removed raw bytes), reserving space for the marker so the result is provably ≤ the limit.
- The packet is built by joining the per-line JSON (`[${serializedLines.join(',')}]`), so each line is serialized **only once** — byte-identical output to `JSON.stringify(array)` with no extra pass.

**Performance** (this branch vs `main`, measured with `main`'s `Logger` and matching config):
- Normal logs: **+0.13 µs/log (+15.6%)** — the irreducible cost of measuring each line's serialized size; sub-millisecond total for a realistic session. Packet output is **byte-identical** to `main` for normal traffic.
- One oversized ~1 MB message: ~18 ms (only fires on logs the server would otherwise reject).

### Fixed Issues
$ https://github.com/Expensify/App/issues/95227

# Tests

- `serializeLineWithinByteLimit`: truncates on **serialized** (escaped) size not raw size; replaces oversized `parameters` with a size marker while keeping the message; bounds lines where both message and parameters are large; handles 6×-expanding control characters.
- `truncateMessageToFitLine`: returns the message unchanged when the line fits; appends the byte marker; reports the correct removed-byte count; never splits astral characters; accounts for JSON escaping; stays within `maxSize` across digit-boundary sizes.
- End-to-end: oversized **message** and oversized **parameters** are both bounded in the sent `logPacket` (whole serialized line ≤ 1,000,000); the debug console still receives the full message.
- The escaped-size math was additionally fuzz-validated against `JSON.stringify` (quotes, control chars, astral pairs, lone surrogates) to confirm the serialized line never exceeds the limit.

**Manual verification:**
1. Checkout this branch, run `npm run build`, and link it into E/App with `npx link publish <path_to_expensify-common>`.
2. In `src/setup/addUtilsToWindow.ts` in E/App, import `Log` and add `window.Log = Log;` inside `addUtilsToWindow`.
3. Run web with `npm run web`.
4. **Under the limit (no truncation):** run `Log.info('A'.repeat(900_000), true);`. In VictoriaLogs, assert the message does **not** end with `...[truncated N bytes]`.
5. **Oversized message (truncated):** run `Log.info('A'.repeat(1_100_000), true);`. Assert the message ends with `...[truncated N bytes]`.
6. **Oversized parameters:** run `Log.info('parameters test', true, {blob: 'A'.repeat(1_100_000)});`. Assert the message is intact (`[info] parameters test`) and `parameters` will include `truncated` and `originalByteSize`.

# QA

N/A

### DIFF
--- a/__tests__/Logger-test.js
+++ b/__tests__/Logger-test.js
@@ -1,5 +1,5 @@
 import {TextEncoder} from 'util';
-import Logger, {truncateToByteSize} from '../lib/Logger';
+import Logger, {truncateMessageToFitLine, enforceLineByteLimit, MAX_LOG_LINE_BYTES} from '../lib/Logger';
 
 const MARKER_REGEX = /\.\.\.\[truncated \d+ bytes\]$/;
 // Independent, UTF-8-correct byte counter for assertions (TextEncoder is the gold standard).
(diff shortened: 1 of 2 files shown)


fabioh8010:

mountiny (approved):
LGTM

PR #929, opened 2026-07-13
[No QA] Remove `sequenceNumber`

s77rt:
Removed functions that are no longer needed/used, these are based on `sequenceNumber` and we are removing that from backend so these function should not be kept.

- Removed `insertIntoCache`
- Made `get` an alias to `getFlatHistory` (they have same outcome, even though the `get` function sends a `offset` param, this param is not used in Web-E)

### Fixed Issues
https://github.com/Expensify/Expensify/issues/589297

# Tests

n/a

# QA

n/a

### DIFF
--- a/lib/ReportHistoryStore.jsx
+++ b/lib/ReportHistoryStore.jsx
@@ -12,6 +12,7 @@ export default function ReportHistoryStore(API, PubSub) {
         throw new Error('Cannot instantiate ReportHistoryStore without API');
     }
 
+
     const store = {
         API,
         cache: {},
--- a/lib/ReportHistoryStore.jsx
+++ b/lib/ReportHistoryStore.jsx
@@ -191,7 +191,7 @@ export default function ReportHistoryStore(API, PubSub) {
             getFromCache(reportID)
                 .done((cachedHistory) => {
                     if (cachedHistory.some(({reportActionID}) => reportActionID === reportAction.reportActionID)) {
-                        mergeHistoryByTimestamp(reportID, [reportAction]);
+                        mergeHistoryByReportActionID(reportID, [reportAction]);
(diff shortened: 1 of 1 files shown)


lib/ReportHistoryStore.jsx
```diff
@@ -183,32 +148,12 @@ export default function ReportHistoryStore(API, PubSub) {
...
-            return promise;
-        },
-
         insertIntoCacheByActionID: (reportID, reportAction) => {
             const promise = new Deferred();
             getFromCache(reportID)
                 .done((cachedHistory) => {
                     if (cachedHistory.some(({reportActionID}) => reportActionID === reportAction.reportActionID)) {
-                        mergeHistoryByTimestamp(reportID, [reportAction]);
+                        mergeHistoryByReportActionID(reportID, [reportAction]);
```
s77rt: This is unrelated but I think it was a mistake, we should dedupe by report action ID (since we are inserting by report action ID) and not the timestamp

s77rt: TODO - revert this

s77rt:
@chiragsalian Please review (for some reason I can't add reviewers in this repo)

s77rt:
The failing tests are probably unrelated, please re-run

stitesExpensify:
@MelvinBot merge main

stitesExpensify (approved):

chiragsalian:
Ah the synk failure. We recently stopped using synk so we've to disable in repos that still use it. I've to push a PR to disable it for expensify-common.

chiragsalian:
@s77rt, we disabled synk but for the check you clear you need to push another commit, or merge with main.

s77rt:
@chiragsalian This is good for merge now

chiragsalian (approved):
LGTM

PR #936, opened 2026-07-27
Pin emnapi transitive dependencies to fix CI on main

chuckdries:
`npm ci` fails on main because the lockfile is missing entries for these dependencies. Not sure why, but they were removed in https://github.com/Expensify/expensify-common/commit/6a57aed17c06d03e497e131592ce600048c775b7 and adding them back fixes it.


roryabraham (approved):

roryabraham:
synk is failing because of plan limits, not a legitimate failure

PR #937, opened 2026-08-03
perf: Optimized ExpensiMark parser for long text

Uzaifm127:
### Explanation of Change

`ExpensiMark.replace()` runs the autolink, bold, and strikethrough regexes against the complete input.

These regexes contain checks that may scan the same text several times while finding a match. This becomes slow when the composer contains a long message, especially on iOS Safari.

This PR first scans the input for possible URL, bold, and strikethrough parts. It then runs the existing ExpensiMark regex only on those smaller parts.

The existing regex still decides whether each part is valid and performs the replacement. This means we are optimizing the existing parser instead of adding another Markdown or URL parser.

For a 14,501-character input, the local App benchmark showed that `ExpensiMark.replace()` went from around 350 ms to 0-1 ms. Typing and editing the same long input remained smooth on Web and iOS Safari.

### Fixed Issues

$ [#95210](https://github.com/Expensify/App/issues/95210)
PROPOSAL: [#95210 (comment)](https://github.com/Expensify/App/issues/95210#issuecomment-4928540531)

### Tests

1. Go to staging.new.expensify.com or open Expensify App
2. Login with any account
3. Go to any chat
4. Paste any text containing markdown (bold, italic and autolink) and 14000+ characters in the composer.
5. Try to write something after pasting.
6. Verify that the typing experience is smooth and it doesn't lag while typing.
7. Verify that the markdown is preserved when we type after 14k text in the composer.

- [x] Verify that no errors appear in the JS console

### Offline tests

Same as Test

### QA Steps

Same as Test

- [x] Verify that no errors appear in the JS console

### PR Author Checklist

- [x] I linked the correct issue in the `### Fixed Issues` section above
- [x] I wrote clear testing steps that cover the changes made in this PR
    - [x] I added steps for local testing in the `Tests` section
    - [x] I added steps for the expected offline behavior in the `Offline steps` section
    - [x] I added steps for Staging and/or Production testing in the `QA steps` section
    - [x] I added steps to cover failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability).
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I ran the tests on **all platforms** & verified they passed on:
    - [x] Android: Native
    - [x] Android: mWeb Chrome
    - [x] iOS: Native
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
- [x] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
- [x] I followed proper code patterns (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
    - [x] I verified that comments were added to code that is not self explanatory
    - [x] I verified that any new or modified comments were clear, correct English, and explained "why" the code was doing something instead of only explaining "what" the code was doing.
    - [x] I verified any copy / text that was added to the app is grammatically correct in English. It adheres to proper capitalization guidelines (note: only the first word of header/labels should be capitalized), and is either coming verbatim from figma or has been approved by marketing (in order to get marketing approval, ask the Bug Zero team member to add the Waiting for copy label to the issue)
- [x] If a new code pattern is added I verified it was agreed to be used by multiple Expensify engineers
- [x] I followed the guidelines as stated in the [Review Guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md)
- [x] I tested other components that can be impacted by my changes (i.e. if the PR modifies a shared library or component like `Avatar`, I verified the components using `Avatar` are working as expected)
- [x] If a new CSS style is added I verified that:
    - [x] A similar style doesn't already exist
    - [x] The style can't be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/styles/utils/index.ts) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(theme.componentBG)`)
- [x] If new assets were added or existing ones were modified, I verified that:
    - [x] The assets are optimized and compressed (for SVG files, run `npm run compress-svg`)
    - [x] The assets load correctly across all supported platforms.
- [x] If the PR modifies code that runs when editing or sending messages, I tested and verified there is no unexpected behavior for all supported markdown - URLs, single line code, code blocks, quotes, headings, bold, strikethrough, and italic.
- [x] If the PR modifies a generic component, I tested and verified that those changes do not break usages of that component in the rest of the App (i.e. if a shared library or component like `Avatar` is modified, I verified that `Avatar` is working as expected in all cases)
- [x] If the PR modifies a component related to any of the existing Storybook stories, I tested and verified all stories for that component are still working as expected.
- [x] If the PR modifies a component or page that can be accessed by a direct deeplink, I verified that the code functions as expected when the deeplink is used - from a logged in and logged out account.
- [x] If the PR modifies the UI (e.g. new buttons, new UI components, changing the padding/spacing/sizing, moving components, etc) or modifies the form input styles:
    - [x] I verified that all the inputs inside a form are aligned with each other.
    - [x] I added `Design` label and/or tagged `@Expensify/design` so the design team can review the changes.
- [x] I added [unit tests](https://github.com/Expensify/App/blob/main/tests/README.md) for any new feature or bug fix in this PR to help automatically prevent regressions in this user flow.
- [x] If the `main` branch was merged into this PR after a review, I tested again and verified the outcome was still expected according to the `Test` steps.

### Screenshots/Videos
<details>
<summary>Android: Native</summary>

https://github.com/user-attachments/assets/92a27caa-b734-4023-a07e-84611c3a5f25

</details>

<details>
<summary>Android: mWeb Chrome</summary>

https://github.com/user-attachments/assets/5fdbb9bf-a2af-4bc2-9a94-7db18615f27a

</details>

<details>
<summary>iOS: Native</summary>

https://github.com/user-attachments/assets/1de502db-0952-4b9a-b1f3-b91fef9eb5f2

</details>

<details>
<summary>iOS: mWeb Safari</summary>

https://github.com/user-attachments/assets/bd02c2e7-15d8-4dd0-8b10-516374ea23b8

</details>

<details>
<summary>MacOS: Chrome / Safari</summary>

https://github.com/user-attachments/assets/c4366c2b-6eac-4f81-946d-c0debb303d3e

</details>

### DIFF
--- a/lib/ExpensiMark.ts
+++ b/lib/ExpensiMark.ts
@@ -80,6 +80,9 @@ const MARKDOWN_VIDEO_REGEX = new RegExp(
     `\\!(?:\\[([^\\][]*(?:\\[[^\\][]*][^\\][]*)*)])?\\(((${UrlPatterns.MARKDOWN_URL_REGEX})\\.(?:${Constants.CONST.VIDEO_EXTENSIONS.join('|')}))\\)(?![^<]*(<\\/pre>|<\\/code>))`,
     'gi',
 );
+const BOLD_MARKDOWN_REGEX =
+    /(?<!<[^>]*)(\b_|\B)\*(?!(?:<\/em))(?![^<]*(?:<\/pre>|<\/code>|<\/a>|<\/video>))((?![\s*])[\s\S]*?[^\s*](?<!\s))\*\B(?![^<]*>)(?![^<]*(<\/pre>|<\/code>|<\/a>|<\/video>))/g;
+const STRIKETHROUGH_MARKDOWN_REGEX = /(?<!<[^>]*)\B~((?![\s~])[\s\S]*?[^\s~](?<!\s))~\B(?![^<]*>)(?![^<]*(<\/pre>|<\/code>|<\/a>|<\/video>))/g;
 
 const SLACK_SPAN_NEW_LINE_TAG = '<span class="c-mrkdwn__br" data-stringify-type="paragraph-break" style="box-sizing: inherit; display: block; height: unset;"></span>';
 
@@ -128,6 +131,228 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
         return text.replace(regexp, (...args) => replacement(extras, ...args));
     }
     return text.replace(regexp, replacement);
+}
+
+function replaceBoldMarkdown(text: string, regexp: RegExp, replacement: Replacement): string {
+    if (!text.includes('*')) {
+        return text;
+    }
+
+    const starPositions = [];
+    const protectedTags = [];
+    const blockedTagNames = new Set(['a', 'code', 'pre', 'video']);
+    let index = 0;
+
+    while (index < text.length) {
+        if (text[index] === '<') {
+            const tagEnd = text.indexOf('>', index + 1);
+            if (tagEnd === -1) {
+                break;
+            }
+
+            const tag = text.slice(index + 1, tagEnd).trim();
+            const isClosingTag = tag.startsWith('/');
+            const tagName = tag.match(/^\/?\s*([a-z][a-z0-9-]*)/i)?.[1]?.toLowerCase();
+            if (tagName && blockedTagNames.has(tagName)) {
+                if (isClosingTag) {
+                    const matchingTagIndex = protectedTags.lastIndexOf(tagName);
+                    if (matchingTagIndex !== -1) {
+                        protectedTags.splice(matchingTagIndex, 1);
+                    }
+                } else if (!tag.endsWith('/')) {
+                    protectedTags.push(tagName);
+                }
+            }
+
+            index = tagEnd + 1;
+            continue;
+        }
+
+        if (text[index] === '*' && protectedTags.length === 0) {
+            starPositions.push(index);
+        }
+        index++;
+    }
+
+    if (starPositions.length < 2) {
+        return text;
+    }
+
+    const isWordCharacter = (character?: string) => {
+        if (!character) {
+            return false;
+        }
+        const code = character.charCodeAt(0);
+        return character === '_' || (code >= 48 && code <= 57) || (code >= 65 && code <= 90) || (code >= 97 && code <= 122);
+    };
+    const canOpen = (position: number) => {
+        const nextCharacter = text[position + 1];
+        if (!nextCharacter || /\s|\*/.test(nextCharacter) || text.startsWith('</em', position + 1)) {
+            return false;
+        }
+
+        const previousCharacter = text[position - 1];
+        if (!isWordCharacter(previousCharacter)) {
+            return true;
+        }
+        return previousCharacter === '_' && !isWordCharacter(text[position - 2]);
+    };
+    const canClose = (position: number) => {
+        const previousCharacter = text[position - 1];
+        return Boolean(previousCharacter && !/\s|\*/.test(previousCharacter) && !isWordCharacter(text[position + 1]));
+    };
+
+    const output = [];
+    const candidateRegex = regexp;
+    let outputStart = 0;
+    let openingPosition;
+
+    for (const starPosition of starPositions) {
+        if (openingPosition === undefined) {
+            openingPosition = canOpen(starPosition) ? starPosition : undefined;
+            continue;
+        }
+        if (!canClose(starPosition)) {
+            continue;
+        }
+
+        const prefixLength = openingPosition > 0 ? 1 : 0;
+        const suffixLength = starPosition + 1 < text.length ? 1 : 0;
+        const candidateStart = openingPosition - prefixLength;
+        const candidateEnd = starPosition + 1 + suffixLength;
+        const candidate = text.slice(candidateStart, candidateEnd);
+        candidateRegex.lastIndex = 0;
+        const candidateMatch = candidateRegex.exec(candidate);
+        if (!candidateMatch) {
+            openingPosition = canOpen(starPosition) ? starPosition : undefined;
+            continue;
+        }
+        candidateRegex.lastIndex = 0;
+        const replacedCandidate = replaceTextWithExtras(candidate, candidateRegex, EXTRAS_DEFAULT, replacement);
+        if (replacedCandidate !== candidate) {
+            const replacedCoreEnd = suffixLength ? replacedCandidate.length - suffixLength : replacedCandidate.length;
+            output.push(text.slice(outputStart, openingPosition));
+            output.push(replacedCandidate.slice(prefixLength, replacedCoreEnd));
+            outputStart = starPosition + 1;
+            openingPosition = undefined;
+            continue;
+        }
+        openingPosition = undefined;
+    }
+
+    if (output.length === 0) {
+        return text;
+    }
+
+    output.push(text.slice(outputStart));
+    return output.join('');
+}
+
+function replaceStrikethroughMarkdown(text: string, regexp: RegExp, replacement: Replacement): string {
+    if (!text.includes('~')) {
+        return text;
+    }
+
+    const tildePositions = [];
+    const protectedTags = [];
+    const blockedTagNames = new Set(['a', 'code', 'pre', 'video']);
+    let index = 0;
+
+    while (index < text.length) {
+        if (text[index] === '<') {
+            const tagEnd = text.indexOf('>', index + 1);
+            if (tagEnd === -1) {
+                break;
+            }
+
+            const tag = text.slice(index + 1, tagEnd).trim();
+            const isClosingTag = tag.startsWith('/');
+            const tagName = tag.match(/^\/?\s*([a-z][a-z0-9-]*)/i)?.[1]?.toLowerCase();
+            if (tagName && blockedTagNames.has(tagName)) {
+                if (isClosingTag) {
+                    const matchingTagIndex = protectedTags.lastIndexOf(tagName);
+                    if (matchingTagIndex !== -1) {
+                        protectedTags.splice(matchingTagIndex, 1);
+                    }
+                } else if (!tag.endsWith('/')) {
+                    protectedTags.push(tagName);
+                }
+            }
+
+            index = tagEnd + 1;
+            continue;
+        }
+
+        if (text[index] === '~' && protectedTags.length === 0) {
+            tildePositions.push(index);
+        }
+        index++;
+    }
+
+    if (tildePositions.length < 2) {
+        return text;
+    }
+
+    const isWordCharacter = (character?: string) => {
+        if (!character) {
+            return false;
+        }
+        const code = character.charCodeAt(0);
+        return character === '_' || (code >= 48 && code <= 57) || (code >= 65 && code <= 90) || (code >= 97 && code <= 122);
+    };
+    const canOpen = (position: number) => {
+        const nextCharacter = text[position + 1];
+        return Boolean(nextCharacter && !/\s|~/.test(nextCharacter));
+    };
+    const canClose = (position: number) => {
+        const previousCharacter = text[position - 1];
+        return Boolean(previousCharacter && !/\s|~/.test(previousCharacter) && !isWordCharacter(text[position + 1]));
+    };
+
+    const output = [];
+    const candidateRegex = regexp;
+    let outputStart = 0;
+    let openingPosition;
+
+    for (const tildePosition of tildePositions) {
+        if (openingPosition === undefined) {
+            openingPosition = canOpen(tildePosition) ? tildePosition : undefined;
+            continue;
+        }
+        if (!canClose(tildePosition)) {
+            continue;
+        }
+
+        const prefixLength = openingPosition > 0 ? 1 : 0;
+        const suffixLength = tildePosition + 1 < text.length ? 1 : 0;
+        const candidateStart = openingPosition - prefixLength;
+        const candidateEnd = tildePosition + 1 + suffixLength;
+        const candidate = text.slice(candidateStart, candidateEnd);
+        candidateRegex.lastIndex = 0;
+        const candidateMatch = candidateRegex.exec(candidate);
+        if (!candidateMatch) {
+            openingPosition = canOpen(tildePosition) ? tildePosition : undefined;
+            continue;
+        }
+        candidateRegex.lastIndex = 0;
+        const replacedCandidate = replaceTextWithExtras(candidate, candidateRegex, EXTRAS_DEFAULT, replacement);
+        if (replacedCandidate !== candidate) {
+            const replacedCoreEnd = suffixLength ? replacedCandidate.length - suffixLength : replacedCandidate.length;
+            output.push(text.slice(outputStart, openingPosition));
+            output.push(replacedCandidate.slice(prefixLength, replacedCoreEnd));
+            outputStart = tildePosition + 1;
+            openingPosition = undefined;
+            continue;
+        }
+        openingPosition = undefined;
+    }
+
+    if (output.length === 0) {
+        return text;
+    }
+
+    output.push(text.slice(outputStart));
+    return output.join('');
 }
 
 /**
@@ -688,7 +913,7 @@ export default class ExpensiMark {
 
                 process: (textToProcess, replacement) => {
                     const regex = new RegExp(`(?![^<]*>|[^<>]*<\\/(?!h1>))([_*~]*?)${UrlPatterns.MARKDOWN_URL_REGEX}\\1(?!((?:(?!<a).)+)?<\\/a>|[^<]*(<\\/pre>|<\\/code>))`, 'gi');
-                    return this.modifyTextForUrlLinks(regex, textToProcess, replacement as ReplacementFn);
+                    return this.modifyTextForUrlLinks(regex, textToProcess, replacement as ReplacementFn, true);
                 },
 
                 replacement: (_extras, _match, g1, g2) => {
@@ -809,7 +1034,7 @@ export default class ExpensiMark {
                 // \B will match everything that \b doesn't, so it works
                 // for * and ~: https://www.rexegg.com/regex-boundaries.html#notb
                 name: 'bold',
-                regex: /(?<!<[^>]*)(\b_|\B)\*(?!(?:<\/em))(?![^<]*(?:<\/pre>|<\/code>|<\/a>|<\/video>))((?![\s*])[\s\S]*?[^\s*](?<!\s))\*\B(?![^<]*>)(?![^<]*(<\/pre>|<\/code>|<\/a>|<\/video>))/g,
+                process: (textToProcess, replacement) => replaceBoldMarkdown(textToProcess, BOLD_MARKDOWN_REGEX, replacement),
                 replacement: (_extras, match, g1, g2) => {
                     if (g1.includes('_')) {
                         return `${g1}<strong>${g2}</strong>`;
@@ -820,7 +1045,7 @@ export default class ExpensiMark {
             },
             {
                 name: 'strikethrough',
-                regex: /(?<!<[^>]*)\B~((?![\s~])[\s\S]*?[^\s~](?<!\s))~\B(?![^<]*>)(?![^<]*(<\/pre>|<\/code>|<\/a>|<\/video>))/g,
+                process: (textToProcess, replacement) => replaceStrikethroughMarkdown(textToProcess, STRIKETHROUGH_MARKDOWN_REGEX, replacement),
                 replacement: (_extras, match, g1) => (g1.includes('</pre>') || containsNonPairTag(g1) ? match : `<del>${g1}</del>`),
             },
             {
@@ -1301,7 +1526,108 @@ export default class ExpensiMark {
     /**
      * Checks matched URLs for validity and replace valid links with html elements
      */
-    modifyTextForUrlLinks(regex: RegExp, textToCheck: string, replacement: ReplacementFn): string {
+    modifyTextForUrlLinks(regex: RegExp, textToCheck: string, replacement: ReplacementFn, shouldScanForUrls = false): string {
+        if (shouldScanForUrls) {
+            const output = [];
+            const candidateRegex = regex;
+            let outputStart = 0;
+            let index = 0;
+            let hostnameRunStart = 0;
+            const isAlphaNumeric = (character: string) => {
+                const code = character.charCodeAt(0);
+                return (code >= 48 && code <= 57) || (code >= 65 && code <= 90) || (code >= 97 && code <= 122);
+            };
+            const isHostnameCharacter = (character: string) => isAlphaNumeric(character) || character === '-' || character === '.';
+            const isSpace = (character: string) => {
+                const code = character.charCodeAt(0);
+                return code <= 32 || code === 160;
+            };
+
+            while (index < textToCheck.length) {
+                if (textToCheck[index] === '<') {
+                    const tagEnd = textToCheck.indexOf('>', index + 1);
+                    if (tagEnd === -1) {
+                        break;
+                    }
+
+                    let closingTag = '';
+                    if (textToCheck.startsWith('<a ', index) || textToCheck.startsWith('<a>', index)) {
+                        closingTag = '</a>';
+                    } else if (textToCheck.startsWith('<pre', index)) {
+                        closingTag = '</pre>';
+                    } else if (textToCheck.startsWith('<code', index)) {
+                        closingTag = '</code>';
+                    }
+
+                    if (closingTag) {
+                        const closingTagStart = textToCheck.indexOf(closingTag, tagEnd + 1);
+                        index = closingTagStart === -1 ? textToCheck.length : closingTagStart + closingTag.length;
+                    } else {
+                        index = tagEnd + 1;
+                    }
+                    hostnameRunStart = index;
+                    continue;
+                }
+
+                if (!isHostnameCharacter(textToCheck[index])) {
+                    hostnameRunStart = index + 1;
+                    index++;
+                    continue;
+                }
+                if (textToCheck[index] !== '.') {
+                    index++;
+                    continue;
+                }
+
+                const hostnameStart = hostnameRunStart;
+                let tldEnd = index + 1;
+                while (tldEnd < textToCheck.length && (isAlphaNumeric(textToCheck[tldEnd]) || textToCheck[tldEnd] === '-')) {
+                    tldEnd++;
+                }
+                if (hostnameStart === index || tldEnd === index + 1) {
+                    index++;
+                    continue;
+                }
+
+                let candidateStart = hostnameStart;
+                const protocol = textToCheck.slice(Math.max(0, hostnameStart - 8), hostnameStart).toLowerCase();
+                if (protocol.endsWith('https://')) {
+                    candidateStart = hostnameStart - 8;
+                } else if (protocol.endsWith('http://')) {
+                    candidateStart = hostnameStart - 7;
+                } else if (protocol.endsWith('ftps://')) {
+                    candidateStart = hostnameStart - 7;
+                } else if (protocol.endsWith('ftp://')) {
+                    candidateStart = hostnameStart - 6;
+                }
+                if (candidateStart > 0 && textToCheck[candidateStart - 1] === '@') {
+                    candidateStart--;
+                }
+                while (candidateStart > 0 && '_*~'.includes(textToCheck[candidateStart - 1])) {
+                    candidateStart--;
+                }
+
+                let candidateEnd = tldEnd;
+                while (candidateEnd < textToCheck.length && !isSpace(textToCheck[candidateEnd]) && textToCheck[candidateEnd] !== '<') {
+                    candidateEnd++;
+                }
+
+                const candidate = textToCheck.slice(candidateStart, candidateEnd);
+                candidateRegex.lastIndex = 0;
+                output.push(textToCheck.slice(outputStart, candidateStart));
+                output.push(this.modifyTextForUrlLinks(candidateRegex, candidate, replacement));
+                outputStart = candidateEnd;
+                index = candidateEnd;
+                hostnameRunStart = candidateEnd;
+            }
+
+            if (output.length === 0) {
+                return textToCheck;
+            }
+            output.push(textToCheck.slice(outputStart));
+            return output.join('');
+        }
+
         let match = regex.exec(textToCheck);
         let replacedText = '';
         let startIndex = 0;
--- a/lib/ExpensiMark.ts
+++ b/lib/ExpensiMark.ts
@@ -192,13 +192,13 @@ function canOpenStrikethroughMarkdown(text: string, position: number): boolean {
     }
 
     const nextCharacter = text[position + 1];
-    return Boolean(nextCharacter && !/\s|~/.test(nextCharacter));
+    return !!nextCharacter && !/\s|~/.test(nextCharacter);
 }
 
 /** Returns whether the marker at this position can close a bold or strikethrough range. */
 function canCloseMarkdown(text: string, position: number, marker: '*' | '~'): boolean {
     const previousCharacter = text[position - 1];
-    return Boolean(previousCharacter && !/\s/.test(previousCharacter) && previousCharacter !== marker && !isWordCharacter(text[position + 1]));
+    return !!previousCharacter && !/\s/.test(previousCharacter) && previousCharacter !== marker && !isWordCharacter(text[position + 1]);
 }
 
 /** Records when scanning enters or leaves <a>, <code>, <pre>, or <video>, then returns the character after the tag. */
@@ -227,7 +227,7 @@ function updateProtectedTagStack(text: string, tagStart: number, protectedTags: 
 
 /** Returns whether the character can be part of a hostname such as example.com. */
 function isHostnameCharacter(character?: string): boolean {
-    return Boolean(character && (isAsciiAlphaNumeric(character) || character === '-' || character === '.'));
+    return !!character && (isAsciiAlphaNumeric(character) || character === '-' || character === '.');
 }
 
 /** Returns whether the character is whitespace that ends a possible URL. */
--- a/lib/ExpensiMark.ts
+++ b/lib/ExpensiMark.ts
@@ -45,12 +45,15 @@ const ASCII_LOWERCASE_END = 'z'.charCodeAt(0);
 const ASCII_WHITESPACE_END = ' '.charCodeAt(0);
 const NON_BREAKING_SPACE_CODE = 160;
 const URL_PROTOCOLS = ['https://', 'http://', 'ftps://', 'ftp://'] as const;
+const URL_CANDIDATE_PREFIX_CHARACTERS = '@_*~';
 const PROTECTED_TAG_NAMES = new Set(['a', 'code', 'pre', 'video']);
 
 type ReplacementFn = (extras: Extras, ...matches: string[]) => string;
 type Replacement = ReplacementFn | string;
 type ProcessFn = (textToProcess: string, replacement: Replacement, shouldKeepRawInput: boolean) => string;
 type UrlCandidate = {start: number; end: number};
+type MarkdownMarker = {position: number; isProtected: boolean};
+type CanOpenMarkdown = (text: string, position: number, isProtected: boolean) => boolean;
 
 type CommonRule = {
     name: string;
@@ -162,7 +165,11 @@ function isWordCharacter(character?: string): boolean {
     return character === '_' || isAsciiAlphaNumeric(character);
 }
 
-function canOpenBoldMarkdown(text: string, position: number): boolean {
+function canOpenBoldMarkdown(text: string, position: number, isProtected: boolean): boolean {
+    if (isProtected) {
+        return false;
+    }
+
     const nextCharacter = text[position + 1];
     if (!nextCharacter || /\s|\*/.test(nextCharacter) || text.startsWith('</em', position + 1)) {
         return false;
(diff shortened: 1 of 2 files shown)


Uzaifm127:

Uzaifm127:
While testing some more parser cases and comparing the result with staging, I found the following two cases and want to confirm whether we should handle them in this PR or separately.

### 1. Non-ASCII domain

For this input:

```text
https://münich.com
```

The behavior is already broken on staging. On staging, `https://m` and `nich.com` become two separate links while `ü` remains plain text. With the optimized parser, the text until `ü` remains plain and only `nich.com` becomes a link.

The ASCII/punycode version works correctly on both staging and dev:

```text
https://xn--mnich-kva.com/
```

Correctly supporting the Unicode version requires converting the hostname to punycode for validation while keeping the original Unicode hostname visible in the message. This looks like a different ExpensiMark issue and should not be mixed with this performance PR because staging is already broken for this case too.

@JS00001 I already have the RCA and a possible solution locally for this issue. Can we cover the Unicode domain support in a follow-up issue?

https://github.com/user-attachments/assets/46e5c2d0-aa09-40f1-acdf-68c913db5678

### 2. URL-looking video label

For this input:

```text
![example.com](https://example.com/video.mp4)
```

I wasn't able to see any broken behavior on real devices because the composer looks correct and the sent message still renders as a video. However, direct parser testing through terminal shows that the output is different.

Existing parser output:

```html
<video ...>example.com</video>
```

Optimized parser output:

**Note:** I only got the broken parser output in terminal, in Web or native I got the correct output.

```html
<video ...><a ...>example.com</a></video>
```

The fix is small. The URL scanner can skip everything between `<video>` and `</video>`, just like it already skips links and code blocks.

Since this parser output difference is introduced by the optimization, I'm not sure whether we should fix this case in the current PR or not because I wasn't get any regression behaviour while testing manually, but can you confirm? @JS00001 @linhvovan29546

I've used this to technically reproduce this case:

```ts
npm run build

node - <<'NODE'
const ExpensiMark = require('./dist/ExpensiMark').default;
const parser = new ExpensiMark();

console.log(
    parser.replace('![example.com](https://example.com/video.mp4)'),
);
NODE
```


https://github.com/user-attachments/assets/89b9b64e-2826-4aec-ba48-0c9c741e29a4


https://github.com/user-attachments/assets/373fa17f-372d-42ea-8f4f-2150bdb4a46f


https://github.com/user-attachments/assets/03dadd11-63da-4c36-a13a-46fb00cc44de

Uzaifm127:
@linhvovan29546

I'm getting a wierd behaviour only on Android native (as per my findings, Its not related to ExpensiMark)

When we paste a long text with having `*Bold*` then we get a delay in typing (Not a lag) or latency but when we do similar thing with ~strikethrough~ or autolink then we don't get any delay.

Also if we paste long text (either plain text or having markdown such as autolink or strikethrough) and then we type Bold, italic, inline code then we don't get this typing delay and the typing seems to be smoother.

We are getting the typing delay when we paste the long text with:

- Bold
- italic
- inline code

We don't get this delay with autolink or strikethrough

I found the root cause for this behaviour and I found `StyleSpan` in Android was causing this behaviour but I'm not really sure.


https://github.com/user-attachments/assets/469bfc32-94a6-4e7e-acd4-31b60479b53b

JS00001:
Yeah if the issue exists on staging, lets do a follow up

lib/ExpensiMark.ts
```diff
@@ -1301,7 +1426,108 @@ export default class ExpensiMark {
...
+    modifyTextForUrlLinks(regex: RegExp, textToCheck: string, replacement: ReplacementFn, shouldScanForUrls = false): string {
+        if (shouldScanForUrls) {
+            const output = [];
+            const candidateRegex = regex;
+            let outputStart = 0;
+            let index = 0;
+            let hostnameRunStart = 0;
+            const isAlphaNumeric = (character: string) => {
+                const code = character.charCodeAt(0);
+                return (code >= 48 && code <= 57) || (code >= 65 && code <= 90) || (code >= 97 && code <= 122);
```
JS00001, line 1438: can we explain this/explain the magic numbers

lib/ExpensiMark.ts
```diff
@@ -130,6 +133,128 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
     return text.replace(regexp, replacement);
 }
 
+function isWordCharacter(character?: string): boolean {
+    if (!character) {
+        return false;
+    }
+    const code = character.charCodeAt(0);
+    return character === '_' || (code >= 48 && code <= 57) || (code >= 65 && code <= 90) || (code >= 97 && code <= 122);
```
JS00001, line 141: same, can we explain the magic numbers

lib/ExpensiMark.ts
```diff
@@ -1301,7 +1426,108 @@ export default class ExpensiMark {
...
-    modifyTextForUrlLinks(regex: RegExp, textToCheck: string, replacement: ReplacementFn): string {
+    modifyTextForUrlLinks(regex: RegExp, textToCheck: string, replacement: ReplacementFn, shouldScanForUrls = false): string {
+        if (shouldScanForUrls) {
+            const output = [];
+            const candidateRegex = regex;
+            let outputStart = 0;
+            let index = 0;
+            let hostnameRunStart = 0;
+            const isAlphaNumeric = (character: string) => {
+                const code = character.charCodeAt(0);
```
JS00001, line 1437: can we reuse this? We do something similar above

lib/ExpensiMark.ts
```diff
@@ -1301,7 +1426,108 @@ export default class ExpensiMark {
...
+                    tldEnd++;
+                }
+                if (hostnameStart === index || tldEnd === index + 1) {
+                    index++;
+                    continue;
+                }
+
+                let candidateStart = hostnameStart;
+                const protocol = textToCheck.slice(Math.max(0, hostnameStart - 8), hostnameStart).toLowerCase();
+                if (protocol.endsWith('https://')) {
```
JS00001, line 1494: Can we calculate the length of the protocol and subtract that instead? why the if blocks

JS00001 (changes requested):

Uzaifm127:
I'll look into the review comments on tomorrow. Thanks

Uzaifm127:
@JS00001 Resolved the review comments:

1. Added constants for the ASCII ranges and reused the same function in `isWordCharacter()` and the URL scanner.
2. Added one list for the supported protocols and used its length instead of hard-coded numbers.

lib/ExpensiMark.ts
Uzaifm127: Fixed. Now we are using the single array of protocols

linhvovan29546:
Reviewing...

linhvovan29546:
@codex review

linhvovan29546:
@codex review

lib/ExpensiMark.ts
```diff
@@ -1301,7 +1443,102 @@ export default class ExpensiMark {
...
+                }
+
+                if (!isHostnameCharacter(textToCheck[index])) {
+                    hostnameRunStart = index + 1;
+                    index++;
+                    continue;
+                }
+                if (textToCheck[index] !== '.') {
+                    index++;
+                    continue;
```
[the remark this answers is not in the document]
linhvovan29546: I'm not sure about the bot's comment here, since I can't reproduce the case Codex mentioned.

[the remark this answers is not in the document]
Uzaifm127: ~I'm currently checking it from my side~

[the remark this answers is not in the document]
Uzaifm127: I'm able to reproduce this case. It isn't reproducible when only pasting `http://localhost:3000/`. But will a user put `localhost` in the composer? 🤔

IMO, we should cover this case.

@linhvovan29546 WDYT?

https://github.com/user-attachments/assets/3a83f55c-a605-4c64-b4ad-17410f3bc113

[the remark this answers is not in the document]
linhvovan29546: I think we should cover this case, since it doesn't happen on the main branch.

lib/ExpensiMark.ts
```diff
@@ -1301,7 +1443,102 @@ export default class ExpensiMark {
...
+                while (tldEnd < textToCheck.length && (isAsciiAlphaNumeric(textToCheck[tldEnd]) || textToCheck[tldEnd] === '-')) {
+                    tldEnd++;
+                }
+                if (hostnameStart === index || tldEnd === index + 1) {
+                    index++;
+                    continue;
+                }
+
+                let candidateStart = hostnameStart;
+                // // Include a protocol such as "https://" so the existing URL regex sees the full URL.
```
linhvovan29546: Minor: doubled comment typo

lib/ExpensiMark.ts
```diff
@@ -130,6 +141,137 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+    }
+    return previousCharacter === '_' && !isWordCharacter(text[position - 2]);
+}
+
+function canOpenStrikethroughMarkdown(text: string, position: number): boolean {
+    const nextCharacter = text[position + 1];
+    return Boolean(nextCharacter && !/\s|~/.test(nextCharacter));
+}
+
+function replaceMarkdownCandidates(text: string, regexp: RegExp, replacement: Replacement, marker: '*' | '~', canOpen: (text: string, position: number) => boolean): string {
```
linhvovan29546: **Duplicated, inconsistent tag-skipping logic.** `replaceMarkdownCandidates` and `modifyTextForUrlLinks`'s new scan branch both reimplement "skip past protected HTML tags," differently:
- `replaceMarkdownCandidates` uses a stack (`protectedTags`), a generic tag-name regex, and a `blockedTagNames` Set (`a`, `code`, `pre`, `video`).
- `modifyTextForUrlLinks` hardcodes three `if/else if` branches for `<a `, `<pre`, `<code`, finds the closing tag via a single `indexOf` (no stack, no nesting support), and doesn't protect `video` at all.

Worth extracting one shared helper (using the stack-based approach) so both call sites can't silently diverge.

lib/ExpensiMark.ts
```diff
@@ -1301,7 +1443,102 @@ export default class ExpensiMark {
     /**
      * Checks matched URLs for validity and replace valid links with html elements
      */
-    modifyTextForUrlLinks(regex: RegExp, textToCheck: string, replacement: ReplacementFn): string {
+    modifyTextForUrlLinks(regex: RegExp, textToCheck: string, replacement: ReplacementFn, shouldScanForUrls = false): string {
+        if (shouldScanForUrls) {
```
linhvovan29546, line 1588: The `shouldScanForUrls` branch in `modifyTextForUrlLinks` is ~80 lines of nested conditionals mutating shared loop state (`index`, `hostnameRunStart`, `outputStart`, `output`) in place. Breaking it into smaller named steps (`skipTag`, `findHostnameEnd`, `extendCandidateBoundaries`) would make each piece independently testable.

lib/ExpensiMark.ts
```diff
@@ -1301,7 +1443,102 @@ export default class ExpensiMark {
...
+            const output = [];
+            const candidateRegex = regex;
+            let outputStart = 0;
+            let index = 0;
+            let hostnameRunStart = 0;
+            const isHostnameCharacter = (character: string) => isAsciiAlphaNumeric(character) || character === '-' || character === '.';
```
linhvovan29546: `isHostnameCharacter` and `isSpace` are defined as closures *inside* `modifyTextForUrlLinks`, recreated on every call (including recursive calls). Everywhere else in this PR, equivalent predicates (`isAsciiAlphaNumeric`, `isWordCharacter`) are correctly hoisted to module scope. Same treatment here would be more consistent and unit-testable.

lib/ExpensiMark.ts
```diff
@@ -1301,7 +1443,102 @@ export default class ExpensiMark {
...
+            const isSpace = (character: string) => {
+                const code = character.charCodeAt(0);
+                return code <= 32 || code === 160;
```
linhvovan29546: 32/160 (space, nbsp) are unexplained, but this PR already introduces ASCII_DIGIT_START etc. as named constants for exactly this. Same treatment here would be consistent.

linhvovan29546 (changes requested):

Uzaifm127:
I'll go through the suggestions today. Thanks

lib/ExpensiMark.ts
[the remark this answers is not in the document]
Uzaifm127: Fixed!

lib/ExpensiMark.ts
Uzaifm127: Removed the entire comment and added self explanatory comments for newly added functions.

Uzaifm127:
@linhvovan29546 I'm still working on suggestion but I have a question from you:

After this [comment](https://github.com/Expensify/expensify-common/pull/937#issuecomment-5251318261), have you used your own codex account to enable codex review in this PR? If so then what GPT model you used to review the PR?

I'm asking this because it'll be better for me in future to review the PR from my side before passing it for review.

linhvovan29546:
> After this https://github.com/Expensify/expensify-common/pull/937#issuecomment-5251318261, have you used your own codex account to enable codex review in this PR? If so then what GPT model you used to review the PR?
I'm asking this because it'll be better for me in future to review the PR from my side before passing it for review.

That asked me to connect GitHub to Codex, so I didn't configure anything. I don't know which GPT model I used.

lib/ExpensiMark.ts
Uzaifm127: Fixed!

lib/ExpensiMark.ts
Uzaifm127: Done!

lib/ExpensiMark.ts
Uzaifm127: Done!

lib/ExpensiMark.ts
Uzaifm127: Done!

Uzaifm127:
@linhvovan29546 I've resolved all the suggestion given by you. You can review now, thanks.

Uzaifm127:
@linhvovan29546 

While reviewing this PR locally, I found and fixed two regressions:

1. A formatted domain after `@`, such as `@*example.com*`, was incorrectly autolinked because the scanner moved across the Markdown marker before checking for `@`. The URL candidate now keeps `@`, `*`, `_`, and `~` together, so the existing URL logic receives the same context as before.

   1. Open any chat.
   2. Paste `@*example.com*` into the composer.
   3. Check that `example.com` is bold and is not shown as a blue link.
   4. Send the message and check the sent message too.

    https://github.com/user-attachments/assets/75b78ff4-151e-45e3-8d89-4e4069192f92

2. Markers inside protected content, such as `` `~` ~strike~ `` or `[~](https://example.com) ~strike~`, were removed from the marker sequence. This could make the scanner select a different Markdown pair than the original parser. The scanner now keeps those markers in order and records whether they are protected, so they can affect pairing without being formatted incorrectly.

   1. Open any chat.
   2. Paste `` `~` ~strike~ `` into the composer.
   3. Check that the first `~` stays inside the code block and only `strike` is strikethrough.
   4. Send the message and check the sent message.
   6. Repeat the same steps with `[~](https://example.com) ~strike~`.
   7. Check that `~` remains the link text and only `strike` is strikethrough.

    https://github.com/user-attachments/assets/887bb902-1bdf-4ba7-b4b8-818253d2ad4e


---

These two suggestions have been fixed in this [commit](https://github.com/Expensify/expensify-common/pull/937/commits/ff829aba2eb6cf1aa1639e2035cd4415aee4797d).

linhvovan29546:
Reviewing again

lib/ExpensiMark.ts
```diff
@@ -130,6 +148,264 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+            markers.push({position: index, isProtected: protectedTags.length > 0});
+        }
+        index++;
+    }
+
+    if (markers.length < 2) {
+        return text;
+    }
+
+    const canClose = (position: number) => {
```
linhvovan29546: NAB: Could you move this function outside of `replaceMarkdownCandidates`?

lib/ExpensiMark.ts
```diff
@@ -130,6 +148,264 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+        }
+
+        // Keep protected markers in order, but remember that they cannot be replaced.
+        if (text[index] === marker) {
+            markers.push({position: index, isProtected: protectedTags.length > 0});
+        }
+        index++;
+    }
+
+    if (markers.length < 2) {
```
linhvovan29546, line 359: Could you please add a comment explaining this?

lib/ExpensiMark.ts
```diff
@@ -130,6 +148,264 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+        const candidate = extendUrlCandidateBoundaries(text, hostnameRunStart, hostnameEnd);
+        candidates.push(candidate);
+        index = candidate.end;
+        hostnameRunStart = candidate.end;
+    }
+
+    return candidates;
+}
+
+function replaceMarkdownCandidates(text: string, regexp: RegExp, replacement: Replacement, marker: '*' | '~', canOpen: CanOpenMarkdown): string {
```
linhvovan29546, line 332: Could you please add a comment explaining this function?

lib/ExpensiMark.ts
```diff
@@ -130,6 +148,264 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+    }
+
+    let candidateEnd = end;
+    while (candidateEnd < text.length && !isUrlBoundarySpace(text[candidateEnd]) && text[candidateEnd] !== '<') {
+        candidateEnd++;
+    }
+    return {start: candidateStart, end: candidateEnd};
+}
+
+// Finds possible URL ranges, skips URL-looking text inside protected tags, and leaves validity to the existing regex.
```
linhvovan29546:
```suggestion
/** Finds possible URL ranges, skips URL-looking text inside protected tags, and leaves validity to the existing regex. */
```

lib/ExpensiMark.ts
```diff
@@ -130,6 +148,264 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+    while (hostnameEnd < text.length && (isAsciiAlphaNumeric(text[hostnameEnd]) || text[hostnameEnd] === '-')) {
+        hostnameEnd++;
+    }
+    if (hostnameStart === dotPosition || hostnameEnd === dotPosition + 1) {
+        return undefined;
+    }
+    return hostnameEnd;
+}
+
+// Expands example.com to include nearby @, *, _, or ~ in any order, plus its path, until whitespace or HTML.
```
linhvovan29546:
```suggestion
/** Expands example.com to include nearby @, *, _, or ~ in any order, plus its path, until whitespace or HTML. */
```

linhvovan29546 (changes requested):
Could you please check and update the comment formatting for all the new functions? Thanks!

linhvovan29546:
## Reviewer Checklist

- [x] I have verified the author checklist is complete (all boxes are checked off).
- [x] I verified the correct issue is linked in the `### Fixed Issues` section above
- [x] I verified testing steps are clear and they cover the changes made in this PR
    - [x] I verified the steps for local testing are in the `Tests` section
    - [x] I verified the steps for Staging and/or Production testing are in the `QA steps` section
    - [x] I verified the steps cover any possible failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [x] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
- [x] I checked that screenshots or videos are included for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [x] I verified that the composer does not automatically focus or open the keyboard on mobile unless explicitly intended. This includes checking that returning the app from the background does not unexpectedly open the keyboard.
- [x] I verified tests pass on **all platforms** & I tested again on:
    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
- [x] If there are any errors in the console that are unrelated to this PR, I either fixed them (preferred) or linked to where I reported them in Slack
- [x] I verified proper code patterns were followed (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
    - [x] I verified that comments were added to code that is not self explanatory
    - [x] I verified that any new or modified comments were clear, correct English, and explained "why" the code was doing something instead of only explaining "what" the code was doing.
    - [x] I verified any copy / text that was added to the app is grammatically correct in English. It adheres to proper capitalization guidelines (note: only the first word of header/labels should be capitalized), and is either coming verbatim from figma or has been approved by marketing (in order to get marketing approval, ask the Bug Zero team member to add the Waiting for copy label to the issue)
- [x] If a new code pattern is added I verified it was agreed to be used by multiple Expensify engineers
- [x] I verified that this PR follows the guidelines as stated in the [Review Guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md)
- [x] I verified other components that can be impacted by these changes have been tested, and I retested again (i.e. if the PR modifies a shared library or component like `Avatar`, I verified the components using `Avatar` have been tested & I retested again)
- [x] If a new component is created I verified that:
    - [x] A similar component doesn't exist in the codebase
    - [x] All props are defined accurately
    - [x] The component has a clear name that is non-ambiguous and the purpose of the component can be inferred from the name alone
    - [x] The only data being stored in the state is data necessary for rendering and nothing else
    - [x] The component has the minimum amount of code necessary for its purpose, and it is broken down into smaller components in order to separate concerns and functions
- [x] If a new CSS style is added I verified that:
    - [x] A similar style doesn't already exist
    - [x] The style can't be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/utils/index.ts) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(theme.componentBG`)
- [x] If the PR modifies code that runs when editing or sending messages, I tested and verified there is no unexpected behavior for all supported markdown - URLs, single line code, code blocks, quotes, headings, bold, strikethrough, and italic.
- [x] If the PR modifies a generic component, I tested and verified that those changes do not break usages of that component in the rest of the App (i.e. if a shared library or component like `Avatar` is modified, I verified that `Avatar` is working as expected in all cases)
- [x] If the PR modifies a component related to any of the existing Storybook stories, I tested and verified all stories for that component are still working as expected.
- [x] If the PR modifies a component or page that can be accessed by a direct deeplink, I verified that the code functions as expected when the deeplink is used - from a logged in and logged out account.
- [x] If the PR modifies the UI (e.g. new buttons, new UI components, changing the padding/spacing/sizing, moving components, etc) or modifies the form input styles:
    - [x] I verified that all the inputs inside a form are aligned with each other.
    - [x] I added `Design` label and/or tagged `@Expensify/design` so the design team can review the changes.
- [x] For any bug fix or new feature in this PR, I verified that sufficient [unit tests](https://github.com/Expensify/App/blob/main/tests/README.md) are included to prevent regressions in this flow.
- [x] If the `main` branch was merged into this PR after a review, I tested again and verified the outcome was still expected according to the `Test` steps.
- [x] I have checked off every checkbox in the PR reviewer checklist, including those that don't apply to this PR.

### Screenshots/Videos
<details>
<summary>Android: HybridApp</summary>



https://github.com/user-attachments/assets/96cf8572-c5a0-4cba-87ab-029bb5b7f39e


</details>

<details>
<summary>Android: mWeb Chrome</summary>



https://github.com/user-attachments/assets/7e96840b-ffc0-451d-826a-0e3ccec2aa1e


</details>

<details>
<summary>iOS: HybridApp</summary>



https://github.com/user-attachments/assets/82bcf60b-6a83-4c8d-9308-c9285a250ea6


</details>

<details>
<summary>iOS: mWeb Safari</summary>



https://github.com/user-attachments/assets/80db1c0c-70fe-47fd-ae89-6ca9c0ac5809


</details>

<details>
<summary>MacOS: Chrome / Safari</summary>



https://github.com/user-attachments/assets/5d3be713-2ea3-4617-83ce-aeaa8f8a606c


</details>

Uzaifm127:
I'll look into the new suggestion/comments tomorrow, thanks.

Uzaifm127:
@linhvovan29546 I've added and refined the comments in the added functions, moved `onClose` outside from `replaceMarkdownCandidates`.

Could you please review, thanks.

linhvovan29546 (approved):

lib/ExpensiMark.ts
```diff
@@ -130,6 +148,271 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+    if (!character) {
+        return false;
+    }
+
+    const code = character.charCodeAt(0);
+    return (
+        (code >= ASCII_DIGIT_START && code <= ASCII_DIGIT_END) ||
+        (code >= ASCII_UPPERCASE_START && code <= ASCII_UPPERCASE_END) ||
+        (code >= ASCII_LOWERCASE_START && code <= ASCII_LOWERCASE_END)
+    );
```
JS00001, line 162: why are we using char codes, why not a whitelist? Is one quicker than the other

lib/ExpensiMark.ts
```diff
@@ -130,6 +148,271 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+            protectedTags.push(tagName);
+        }
+    }
+
+    return tagEnd + 1;
+}
+
+/** Returns whether the character can be part of a hostname such as example.com. */
+function isHostnameCharacter(character?: string): boolean {
+    return Boolean(character && (isAsciiAlphaNumeric(character) || character === '-' || character === '.'));
```
JS00001: do we need the Boolean()

lib/ExpensiMark.ts
```diff
@@ -130,6 +148,271 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+}
+
+/** Returns whether the character is whitespace that ends a possible URL. */
+function isUrlBoundarySpace(character: string): boolean {
```
JS00001, line 234: why do we need to use char codes for this?

lib/ExpensiMark.ts
```diff
@@ -1301,7 +1584,29 @@ export default class ExpensiMark {
     /**
      * Checks matched URLs for validity and replace valid links with html elements
      */
-    modifyTextForUrlLinks(regex: RegExp, textToCheck: string, replacement: ReplacementFn): string {
+    modifyTextForUrlLinks(regex: RegExp, textToCheck: string, replacement: ReplacementFn, shouldScanForUrls = false): string {
+        if (shouldScanForUrls) {
+            const candidates = findUrlCandidates(textToCheck);
+            if (candidates.length === 0) {
```
JS00001, line 1590: Can we also use a `<2` check here?

JS00001 (changes requested):

Uzaifm127:
~I'll go through each suggestion on Friday (today).~

Looking into the suggestions...

lib/ExpensiMark.ts
Uzaifm127: If you meant this from whitelist: `'0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'` then:

The reason for using char codes was to check the ASCII ranges directly

I've choosen the char codes because it felt straighforward and more direct IMO according to what we need to do. Since we are checking a character whether it is in ASCII range or not so it felt more direct instead of looking into a fixed whitelist.

---

After going through further in this question, I found several advantages too of using `.charCodeAt` here instead of `.includes`:

1. We are using`charCodeAt()` to get the char code of the boundary characters and then doing the comparison, so it will have `O(1)` time complexity but `.includes` has `O(n)`.
2. I've ran the following code to test the benchmark in devtool and got the following result:

### Code 1:

```ts
(() => {
    const digitStart = '0'.charCodeAt(0);
    const digitEnd = '9'.charCodeAt(0);
    const uppercaseStart = 'A'.charCodeAt(0);
    const uppercaseEnd = 'Z'.charCodeAt(0);
    const lowercaseStart = 'a'.charCodeAt(0);
    const lowercaseEnd = 'z'.charCodeAt(0);

    const whitelist = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';

    function usingCharCodes(character) {
        if (!character) {
            return false;
        }

        const code = character.charCodeAt(0);

        return (
            (code >= digitStart && code <= digitEnd) ||
            (code >= uppercaseStart && code <= uppercaseEnd) ||
            (code >= lowercaseStart && code <= lowercaseEnd)
        );
    }

    function usingWhitelist(character) {
        return !!character && whitelist.includes(character);
    }

    const samples = 'aZ7-._@ é中'.repeat(1500);

    function benchmark(fn) {
        let matched = 0;
        const start = performance.now();

        for (let round = 0; round < 200; round++) {
            for (const character of samples) {
                if (fn(character)) {
                    matched++;
                }
            }
        }

        return {
            timeMs: performance.now() - start,
            matched,
        };
    }

    console.table({
        charCodes: benchmark(usingCharCodes),
        whitelist: benchmark(usingWhitelist),
    });
})();
```

image

### Code 2:

```ts
(() => {
    const digitStart = '0'.charCodeAt(0);
    const digitEnd = '9'.charCodeAt(0);
    const uppercaseStart = 'A'.charCodeAt(0);
    const uppercaseEnd = 'Z'.charCodeAt(0);
    const lowercaseStart = 'a'.charCodeAt(0);
    const lowercaseEnd = 'z'.charCodeAt(0);

    const whitelist = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';

    function usingCharCodes(character) {
        if (!character) {
            return false;
        }

        const code = character.charCodeAt(0);

        return (
            (code >= digitStart && code <= digitEnd) ||
            (code >= uppercaseStart && code <= uppercaseEnd) ||
            (code >= lowercaseStart && code <= lowercaseEnd)
        );
    }

    function usingWhitelist(character) {
        return !!character && whitelist.includes(character);
    }

    const samples = Array.from({length: 15000}, (_, index) => {
        const values = ['a', 'Z', '7', '-', '.', '_', '@', ' ', 'é', '中'];
        return values[index % values.length];
    });

    function measure(fn) {
        let matched = 0;
        const start = performance.now();

        for (let round = 0; round < 200; round++) {
            for (const sample of samples) {
                if (fn(sample)) {
                    matched++;
                }
            }
        }

        return {
            timeMs: performance.now() - start,
            matched,
        };
    }

    // Warm up both functions before measuring them.
    measure(usingCharCodes);
    measure(usingWhitelist);

    const charCodeResults = [];
    const whitelistResults = [];

    for (let run = 0; run < 5; run++) {
        charCodeResults.push(measure(usingCharCodes).timeMs);
        whitelistResults.push(measure(usingWhitelist).timeMs);
    }

    function median(values) {
        return [...values].sort((a, b) => a - b)[Math.floor(values.length / 2)];
    }

    console.table({
        charCodeMedianMs: median(charCodeResults),
        whitelistMedianMs: median(whitelistResults),
        charCodeRuns: charCodeResults,
        whitelistRuns: whitelistResults,
    });
})();
```

Screenshot 2026-08-21 at 11 46 38 AM

lib/ExpensiMark.ts
Uzaifm127: Good catch!

Removed the `Boolean` function

lib/ExpensiMark.ts
Uzaifm127: Same reason as [this](https://github.com/Expensify/expensify-common/pull/937#discussion_r3827756964).

image

This check runs for every character while finding URL boundaries. As per the above logs result, the char code check took `47.4 ms`, while the reused `/\s/` regex took `151.5 ms`, so the regex was about 3 times slower. Since the parser is performance-sensitive, we should keep using `.charCodeAt()`.

The second reason is that we want to preserve the current boundary behavior. The current check treats ASCII characters up to code `32` and the non-breaking space (`160`) as URL boundaries. `/\s/.test(character)` also matches additional Unicode whitespace characters, such as `U+2003`, which would change the current behavior.

lib/ExpensiMark.ts
Uzaifm127: No we can't use `candidates.length < 2` here because:

For bold and strikethrough, we need at least two marker positions to create a range. For example, `*Bold*` has two `*` characters around `Bold`, so both markers are needed before we can format the text.

This is different for URLs because one URL is already a complete candidate. For example, `https://example.com` produces one URL candidate, and that single candidate must still be processed and converted into a link.

Therefore, `candidates.length === 0` is correct here. It returns the original text only when no possible URL candidate was found. Changing it to `candidates.length < 2` would prevent a single URL from being autolinked.

Uzaifm127:
@JS00001 I've replied to all the changes requested by you.

Please don't merge this PR as I'm doing the testing regarding [this](https://github.com/Expensify/expensify-common/pull/937#issuecomment-5214021717) Android slowdown (I'm 100% sure this is a different issue and should also be covered in a follow up but I'll provide the conclusion) soon.

TY!

JS00001:
lmk when this is ready please

JS00001:
Any updates?

Uzaifm127:
~Will finalize by EOD~

Uzaifm127:
@JS00001 

All clear from my side now because the Android typing delay seems to be caused by other reason (which we will find out, though I tried to find the reason for that but it needs deeper investigation). It isn't related to `ExpensiMark` because I also got that delay when I pasted almost 14400 characters text without markdown.

We will figure it out in a follow PR/issue, for now I don't think it should block the parser optimization for other platforms.

Sorry for the delay, since I was also doing the codex review locally for this PR, codex gave some regressions but when I tested manually then they weren't reproducible.

TY!

JS00001:
@Uzaifm127 can you take a look at @linhvovan29546 's comments please

Uzaifm127:
@JS00001 We already addressed all comments by @linhvovan29546 

I forgot to resolve those comments, my bad

lib/ExpensiMark.ts
```diff
@@ -35,10 +35,25 @@ type Extras = {
...
+// These constants represent the ASCII ranges for digits (0-9) and letters (A-Z, a-z).
+const ASCII_DIGIT_START = '0'.charCodeAt(0);
+const ASCII_DIGIT_END = '9'.charCodeAt(0);
+const ASCII_UPPERCASE_START = 'A'.charCodeAt(0);
+const ASCII_UPPERCASE_END = 'Z'.charCodeAt(0);
+const ASCII_LOWERCASE_START = 'a'.charCodeAt(0);
+const ASCII_LOWERCASE_END = 'z'.charCodeAt(0);
+const ASCII_WHITESPACE_END = ' '.charCodeAt(0);
+const NON_BREAKING_SPACE_CODE = 160;
+const URL_PROTOCOLS = ['https://', 'http://', 'ftps://', 'ftp://'] as const;
```
JS00001, line 47: Why do we need ftp or ftps? Is that really a real use-case?

lib/ExpensiMark.ts
```diff
@@ -130,6 +148,271 @@ function replaceTextWithExtras(text: string, regexp: RegExp, extras: Extras, rep
...
+/** Returns whether the character is whitespace that ends a possible URL. */
+function isUrlBoundarySpace(character: string): boolean {
+    const code = character.charCodeAt(0);
+    return code <= ASCII_WHITESPACE_END || code === NON_BREAKING_SPACE_CODE;
+}
+
+/** Returns the supported URL protocol that starts at this position, if one exists. */
+function getProtocolAt(text: string, position: number) {
+    const firstCharacter = text[position]?.toLowerCase();
+    if (firstCharacter !== 'h' && firstCharacter !== 'f') {
```
JS00001, line 242: same, do we really need ftp?

JS00001 (changes requested):
one more question

lib/ExpensiMark.ts
Uzaifm127: We allow ftp/ftps in staging so I follow the staging and kept the ftp/ftps.

I compared the new optimized parser with existing parser in staging in terms of parsing, I've kept the staging parsing abilities while optimizing for performance.


https://github.com/user-attachments/assets/f644c447-951c-4e6b-aba4-eb628da077af

lib/ExpensiMark.ts
Uzaifm127: Same, we allow ftp/ftps in staging so kept the same behaviour.

image

JS00001 (approved):

JS00001:
@Uzaifm127 lets create a PR to bump the version in e/app please

Uzaifm127:
I'll raise a bump PR in App soon. Thanks

PR #939, opened 2026-08-25
Enforce Node 24.19.0 / npm 11.10.0 via devEngines and engine-strict

NicolasBonet:
This enforces the Node/npm toolchain for local development and CI so the upcoming npm `min-release-age` supply-chain cooldown cannot be silently bypassed. The `min-release-age` setting requires npm >= 11.10.0 and is silently ignored by older npm versions; Node 24.11.0's bundled npm 11.6.x is below that floor. This PR bumps `.nvmrc` to Node 24.19.0 (same 24.x LTS line, which bundles npm 11.17.0) and declares the floors in `devEngines` (`node >=24.19.0`, `npm >=11.10.0`, both with `onFail: "error"`) rather than `engines`: because this package is published to npm and consumed by repos with `engine-strict` enabled (e.g. App), an `engines` field would impose our repo-tooling floors on every consumer install, breaking consumers on older-but-fine toolchains. `devEngines` is checked only for the root project and never when the package is installed as a dependency, so it enforces the floor for people working in this repo without constraining consumers. `.npmrc` with `engine-strict=true` is also added so any future `engines` declarations in this repo's dependency tree are respected. On npm >= 10.9 a toolchain below the floor fails with `EBADDEVENGINES`; older npm ignores `devEngines` silently, which is acceptable since `.nvmrc` and CI pin the toolchain. All CI workflows already read `node-version-file: .nvmrc`, so no workflow changes were needed.

### Fixed Issues
For https://github.com/Expensify/Expensify/issues/675045

# Tests
1. No unit/integration or autoQA test changes; this is a toolchain configuration change, covered by the existing CI workflows (lint, test, typecheck) which read `.nvmrc` and will now run on Node 24.19.0.
1. Verified `node --version` is v24.19.0 and `npm --version` is 11.17.0 under the new toolchain, then ran `npm install` — completed successfully; the lockfile no longer carries an `engines` mirror in the root entry (npm does not mirror `devEngines`), and the remaining lockfile diff vs main is minimal peer-flag normalization from the newer npm, which is committed.
1. Ran `npm ci` in a clean checkout under Node 24.19.0 / npm 11.17.0 — completed successfully (exit 0).
1. Verified `devEngines` enforcement fires: temporarily set the node floor to `>=99` and ran `npm install --dry-run` under Node 24.19.0 / npm 11.17.0 — failed with `EBADDEVENGINES` (`Invalid semver version ">=99" does not match "v24.19.0" for "runtime"`, exit code 1), then restored the real floor.
1. Confirmed behavior under the system default old toolchain (Node v20.20.0 / npm 10.8.2): npm < 10.9 ignores `devEngines` silently, so `npm install --dry-run` succeeds there — acceptable because `.nvmrc` and CI pin Node 24.19.0, whose bundled npm enforces the floor.

# QA
1. None — internal toolchain configuration change with no runtime behavior impact.
1. No regression areas; published package contents are unchanged, and no `engines` constraint is imposed on consumers.


NicolasBonet:

deetergp (approved):

